PR #454, opened 2014-12-17
Add /usr/local/bin/ to PATH for mvn from homebrew

brunobowden:
Feel free to solve this in an alternative manner if something else is more appropriate


tomball:
This delta finally merged -- the issue was that the JreEmulation project needed an overhaul first.

brunobowden:
@tomball - did you want to add "/opt/local/bin" as well since you mentioned it before?

tomball:
Already done:

grep usr/local/bin
third_party/java_src/j2objc/jre_emul/JreEmulation.xcodeproj/project.pbxproj
buildArgumentsString = "-c \"export
PATH=$PATH:/usr/local/bin:/opt/local/bin; if [ x$ACTION == xclean ]; then
make clean; else make -j4 dist; fi\"";

brunobowden:
Wonderful

PR #457, opened 2014-12-19
Match Java / iOS log levels + Travis-CI build status

brunobowden:


tomball:
I finally updated the script permissions (internally, soon to be pushed), and Travis-CI is running. So I think this pull request is finally complete.

PR #461, opened 2014-12-21
Add targets to allow running individual tests.

harrycheung:
For example:
```
make translate.FunctionizerTest
```


tomball:
We'll leave it as is, to encourage Eclipse debugging.

PR #463, opened 2014-12-27
Translate Java member variables into Objective-C

harrycheung:
properties with Property annotation. For example:
```
class Foo {
    @Property("copy, nonatomic") protected String bar;
}
```
translates to:
```
@interface Foo : NSObject {
}

@property (copy, nonatomic) NSString *bar;

@end

@implementation MyClass

@synthesize bar = bar_;

@end
```


kstanger:
@harrycheung, @tomball asked me to take a look at these changes and I think it's a great addition.

I recently made some changes that will require a bit more work to merge your changes. All of the type generation logic has been extracted into new classes. Your changes to ObjectiveCImplementationGenerator will need to be moved to TypeImplementationGenerator. Your changes to ObjectiveCSourceFileGenerator should be moved to TypeDeclarationGenerator. Hopefully this is mostly a matter of copying and pasting your code.

harrycheung:
Thanks for taking a look. I'll put your comments in when I get a chance.

PR #464, opened 2014-12-29
Adding new line after "error:" and "warning:".

yixiang:
Adding a new line so as to preserve the original message format. If the original message happens to start with "FILE:LINE: ", Xcode will be able to pick it up. For java files added to an Xcode project and configured to translate as part of the build process (just like https://github.com/tomball/j2objc-sample-reversi), we will be able to click on the error message, and it will bring us to the line with error. That make it easy to code to fix java files within Xcode.

However the new line between "error:" and the remaining message may not be desirable in some other cases. Merge with caution.

The perfect solution to this will be having error message and line number coming in as separate parameters, but that requires a much large scope change.


yixiang:
canceling pull request. submitting internally.

PR #466, opened 2015-01-06
Whoops, missing pointer '*' in multi-catch exception translations.

harrycheung:
Not sure how this got missed. Maybe no one uses it. Unit test was wrong too.


tomball:
Thanks for the patch; it's merged in internally, and should be pushed publicly later this week.

You're right about multi-catch not being used, at least by the code being translated. My guess is this is because GWT and Android didn't adopt Java 7 right away, so developers targeting multiple platforms had to avoid using Java 7 language features.

tomball:
Now available in public repository.

PR #469, opened 2015-01-09
Renamed IOSClass.classWithClass and classWithProtocol to IOSClass.classF...

### LINKED ISSUE #468: In Swift, IOSClass doesn't behave correctly.
A little background first: When Swift imports Objective-C code, it changes constructor signatures to a simpler form. For example, in Objective-C, this constructor call:
```
[[Person alloc] initWithFirstName:@"John" lastName:@"Doe"];
```
becomes this in Swift:
```
Person(firstName: "John", lastName: "Doe")
```

Keeping that in mind, let's look at IOSClass. IOSClass has a few static methods for creating IOSClass instances.
```
// IOSClass Getters.
+ (IOSClass *)classWithClass:(Class)cls;
+ (IOSClass *)classWithProtocol:(Protocol *)protocol;
+ (IOSClass *)arrayClassWithComponentType:(IOSClass *)componentType;
+ (IOSClass *)classForIosName:(NSString *)iosName;
+ (IOSClass *)primitiveClassForChar:(unichar)c;
```
Here's the problem. In Swift, Xcode thinks these static methods are actual constructors. You would expect the call to look like:
```
IOSClass.classWithClass(SomeClass.self)
```
but instead, code sense autocompletes to this:
```
IOSClass(class: SomeClass.self)
```
This looks wrong already since that is a constructor call which doesn't exist. Try compiling and you get a  syntax error.

I suspect this is an Xcode bug, but I found a quick fix by renaming 'classWithClass' to 'getClassWithClass'. Considering this will affect a lot of lines of code, I'm posing the solution here first. All tests pass with the renaming.

harrycheung:
...romClass and classFromProtocol.

Swift maps constructors out of class methods that follow a factory naming scheme. This becomes a problem since "WithClass" maps to "class" (and "protocol") which is a reserved word.

classWithClass and classWithProtocol have deprecated messages to maintain backwards compatiblity.

Fixes #468.

### DIFF
--- a/jre_emul/Classes/IOSArrayClass.m
+++ b/jre_emul/Classes/IOSArrayClass.m
@@ -72,7 +72,7 @@
 
 - (IOSObjectArray *)getInterfacesWithArrayType:(IOSClass *)arrayType {
   return [IOSObjectArray arrayWithObjects:(id[]){
-        [IOSClass classWithProtocol:@protocol(JavaIoSerializable)]
+        [IOSClass classFromProtocol:@protocol(JavaIoSerializable)]
       } count:1 type:[IOSClass getClass]];
 }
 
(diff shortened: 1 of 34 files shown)


tomball:
Merged into internal repository, minus the deprecated attributes. We need to update some of our client code to use the new methods, so their builds don't break due to deprecated warnings.

PR #481, opened 2015-01-19
Support not following HTTP redirects

### LINKED ISSUE #387: IosHttpUrlConnection does not respect instanceFollowRedirects (when false)
The `HttpUrlConnection` class has a field `instanceFollowRedirects` that specifies if a redirect should be followed automatically or not.

The `IosHttpUrlConnection` implementation always follows redirects, and thus it does not respect the value of `instanceFollowRedirects`. I have tried to implement this functionality but have failed, and therefore I create this issue instead.

I have found that a [NSURLConnectionDataDelegate](https://developer.apple.com/library/mac/documentation/Foundation/Reference/NSURLConnectionDataDelegate_protocol/Reference/Reference.html) can be used and the method [connection:willSendRequest:redirectResponse:](https://developer.apple.com/library/mac/documentation/Foundation/Reference/NSURLConnectionDataDelegate_protocol/Reference/Reference.html#//apple_ref/occ/intfm/NSURLConnectionDataDelegate/connection:willSendRequest:redirectResponse:) should be implemented. The method should return `nil` to specify that a redirect should not be followed, or it should return the supplied redirected request if the redirect should be followed. It is therefore the ideal candidate for implementing the functionality required for `instanceFollowRedirects`.

However, currently `IosHttpUrlConnection` uses `[NSURLConnection sendSynchronousRequest: ...]` which does not support delegates. I was unable in my attempt to rewrite it to use asynchronous loading, which does support the delegate. I hope my research can be used, and someone with more knowledge of threading can implement my proposed changes.

This issue is blocking me from performing further experimentation with `IosHttpUrlConnection`.

foens:
This update fixes #387:
 * Respect the field value of instanceFollowRedirects
 * Update the URL if the request was redirected
 * Throw SocetTimeoutException on timeouts

### DIFF
--- a/jre_emul/Classes/com/google/j2objc/net/IosHttpURLConnection.java
+++ b/jre_emul/Classes/com/google/j2objc/net/IosHttpURLConnection.java
@@ -43,6 +43,7 @@ import java.util.Map;
 #include "java/net/ConnectException.h"
 #include "java/net/MalformedURLException.h"
 #include "java/net/UnknownHostException.h"
+#include "java/net/SocketTimeoutException.h"
 ]-*/
 
 /**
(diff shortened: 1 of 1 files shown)


tomball:
Merged, and will go out to public repository the next time we sync. Thank you very much for fixing this!

foens:
Hi again Tom.

Does the j2objc library use ARC? I have found that my code has memory leaks, and it seems I have compiled j2objc without ARC. Is this the default? In any case, I can provide another pull request which uses the RELEASE_ and AUTORELEASE macroes, thus avoiding memory leaks.

tomball:
Yes, we compile our libraries without ARC. ARC is great for general
application code since it is retains and releases conservatively (as in,
sometimes more than it needs to), but can hurt performance-critical code.
We found it made a big difference for array and collections' performance.

foens:
Okay. I will provide a fix then.

PR #488, opened 2015-01-24
Add podspec

tamird:
Inspired by the work done in https://github.com/goodow/j2objc but uses cocoapods facilities to download the release rather than shelling out. @tomball /cc @goodow


tamird:
I've signed the CLA.

PR #489, opened 2015-01-28
Fix memory leak

foens:
This fixes the memory leak. Please note, that the first commit is the same as the pull request merged earlier, #481.


foens:
@tomball, I can see that you have already pulled in the changes. Please confirm and close this pull request.

tomball:
Yes, I merged it yesterday, prior to releasing 0.9.6. Thanks again for your fix! The only thing I changed was the indentation, since we have a 100 column maximum.

PR #501, opened 2015-02-26
Questions section now points to StackOverflow

brunobowden:
@tomball - first attempt on pull request for documentation


tomball:
Pulled. Now I need to check whether that screwed up our internal->external repository syncing script.

brunobowden:
It would be nicer if it was a wiki style interface but with a higher
barrier like this, at least you'll only get quality contributions.

Once GitHub comes out with a better workflow, we should ponder if we want
to do something else.

PR #544, opened 2015-06-02
JreStrcat NsNull

Lukasz2891:
JreStrcat: There is a problem when the str as String is null in Java, it has NSNull class instead of being nil.
Try to use toString() method in your custom java object with any null field and then display it in ObjC.

### DIFF
--- a/jre_emul/Classes/JreEmulation.m
+++ b/jre_emul/Classes/JreEmulation.m
@@ -120,7 +120,8 @@ FOUNDATION_EXPORT NSString *JreStrcat(const char *pTypes, ...) {
       case '$':
         {
           NSString *str = va_arg(va, NSString *);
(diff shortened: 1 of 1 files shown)


Lukasz2891:
I signed it!

kstanger:
JreStrcat should not have to handle NSNull values. NSNull is not assignable to NSString and is never instantiated in J2ObjC translated code.
I would recommend making sure that you are not calling into translated code with any NSNull values, since you are likely to come across more issues than just string concatenation.

tomball:
I just verified that [[NSNull null] description] works correctly when used in a JreStrcat expression, using this test:

```java
class Hello {
  static native Object who() /*-[
    return [NSNull null];
  ]-*/;
  public static void main(String[] args) {
    System.out.println("hello " + who());
  }
}
```
It prints: `hello <null>`

Issue #483, opened 2015-01-20
AsyncTask support


OMFOm:
Does j2objc include support for Android thread operations? I'm using the AsyncTask class but I'm getting the error The import android.os.AsyncTask cannot be resolved.

If it's not supported what should I use as an alternative solution?

tomball:
I'm afraid the first sentence in the javadoc for AsyncTask is: "AsyncTask
enables proper and easy use of the UI thread. " J2ObjC doesn't support UI
code by design. AsyncTask is tightly integrated with the Android OS, with
references to Looper, Message, Parcel, etc.

However, if all you want to do is run background tasks,  you can post
FutureTask jobs to a java.util.concurrent.ThreadPoolExecutor. The
ThreadPoolExecutor gives you the async support, while the FutureTask
provides the ability to cancel the task and get execution status. Android
supports that API, so the code will be portable.

PR #545, opened 2015-06-05
Added android.os.AsyncTask.

theoriginalgri:
- Sources based on https://github.com/android/platform_frameworks_base/blob/master/core/java/android/os/AsyncTask.java
- InternalHandler class removed and calls from postResult and publishProgress are made using dispatch_async instead of Handler
- Only commented-out line is 284: Process.setThreadPriority does not seem to be available here

I am not sure whether you would like a port of this class since #483 was rejected. Anyway porting was not a big task and was the easiest way for getting my application to work since I did not find a way to get FutureTask's done() method to return in the main thread as AsyncTask does.

Feel free to accept, comment or reject this pull request. Great project!

### DIFF
--- a/jre_emul/Makefile
+++ b/jre_emul/Makefile
@@ -81,6 +81,7 @@ JAR_ONLY_SOURCES = \
 # Java sources to be translated normally.
 JAVA_SOURCES = \
 	android/os/SystemClock.java \
+	android/os/AsyncTask.java \
 	android/test/suitebuilder/annotation/LargeTest.java \
 	android/test/suitebuilder/annotation/MediumTest.java \
 	android/test/suitebuilder/annotation/SmallTest.java \
--- a/jre_emul/android/frameworks/base/core/java/android/os/AsyncTask.java
+++ b/jre_emul/android/frameworks/base/core/java/android/os/AsyncTask.java
@@ -1,0 +1,640 @@
+/*
+ * Copyright (C) 2008 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*-[
+#include <dispatch/dispatch.h>
+]-*/
+
+package android.os;
+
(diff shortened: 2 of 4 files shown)


theoriginalgri:
I signed it!

tomball:
Thanks, this class looks very useful! How did you test it? We're always looking for tests. If you just have something minimal, you can attach it to this issue and I'll turn it into a JUnit test.

theoriginalgri:
I've also written a unit test but it always fails since dispatch_async does not seem to perform the same on OSX and iOS. You can find it here: https://gist.github.com/theoriginalgri/c94f5e5348e6070dce6d

I've used the "Waiter" class from https://github.com/jhalterman/concurrentunit which may also be a good addition to the junit included in j2objc itself.

When using the AsyncTask class inside an app running on iOS simulator or on a real device, the three functions mentioned as not being called in the unit tests really get invoked - that's how I tested it.

theoriginalgri:
I added the test but did not get it working here since I did not get the unittests to run inside a the simulator. Maybe your travis has more luck on this.

To test again, I wrote a little test application which implemented AndroidOsAsyncTask and ran it in the simulator where all callbacks worked and were in the correct threads (checked this using the debugger):
```
Simulator session started with process 4060
2015-06-06 17:14:22.839 TestApp[4060:321764] onPreExecute
2015-06-06 17:14:22.840 TestApp[4060:321774] doInBackgroundWithNSObjectArray: (null)
2015-06-06 17:14:23.005 TestApp[4060:321764] onProgressUpdateWithNSObjectArray: (null)
2015-06-06 17:14:23.005 TestApp[4060:321764] onPostExecuteWithId: hello world
```

tomball:
Thanks for the test, it looks great. I'll get it working on the simulator before committing.

theoriginalgri:
Just modified the dispatch_async blocks to use a __block self pointer in this last patch.

tomball:
Who is Jonathan Halterman? He's listed in the `@author` tag in the Waiter class in AsyncTaskTest.java. I'll add AsyncTask.java now, but defer adding the test until ownership is cleared up.

theoriginalgri:
He is @jhalterman. I just included the Waiter class inline since it's not available otherwise using j2objc sources.
The repo for the tests is https://github.com/jhalterman as written above.

jhalterman:
`Waiter` is Apache 2 licensed as part of [ConcurrentUnit](https://github.com/jhalterman/concurrentunit) and can be used according to the usual apache 2 licensing rules. I'd think it might be better to pull in concurrentunit as a build dependency rather than copy the code, but whatever works.

tomball:
Thanks, Jonathan. ConcurrentUnit looks very useful, its source is available via Maven and our projects share the same license, so I'll add it as a build dependency as you suggest.

tomball:
Added AsyncTask.java to build. AsyncTaskTest.java wasn't, because it includes source from another library; I plan on using it to write a similar test that doesn't.

PR #555, opened 2015-06-24
Fix the incorrect syscall in Posix.preadBytes

### LINKED ISSUE #554: Posix.preadBytes called pwrite64 in the native method
`libcore.io.Posix.preadBytes` calls `pwrite64` instead of `pread64` in its native method implementation. I discovered this bug when I was reading from a `FileChannel` from a random access file.

I've made a fully reproducible gist [here](https://gist.github.com/lukhnos/5bc1749cd8ed6507e62c). In essence, this snippet:

```java
ByteBuffer buf = ByteBuffer.allocate(bufLength);
RandomAccessFile raf = new RandomAccessFile(dataFile, "r");
FileChannel channel = raf.getChannel();
int bytesRead = channel.read(buf, 0);
channel.close();
```

Would throw this exception:

```
java.io.IOException: pread failed: EBADF (Bad file descriptor)
    at 0x000000010a1cc4f0 libcore.io.ErrnoException.rethrowAsIOException() + 38
    at 0x000000010a11e749 java.nio.FileChannelImpl.readImpl() + 473
    at 0x000000010a08828d PreadBug.main() + 445
    at 0x000000010a0880bd PreadBug.main() + 29
    at 0x000000010a20664e java.lang.reflect.Method.invoke() + 634
```

Also, if the random access file is opened with mode `"rw"`, the code will finish quietly, causing the file to be overwritten with the content in the buffer, since `pwrite64` is actually called to read from the channel.

lukhnos:
This pull request fixes #554: `libcore.io.Posix.preadBytes` calls `pwrite64` in its native method implementation when `pread64` should be called instead.

A reproducible gist is provided [here](https://gist.github.com/lukhnos/5bc1749cd8ed6507e62c).

After applying the patch, run `make jre_emul_dist` and rebuild the transpiled project. I used the above-mentioned gist to make sure that the bug would be fixed this way.

### DIFF
--- a/jre_emul/android/libcore/luni/src/main/java/libcore/io/Posix.java
+++ b/jre_emul/android/libcore/luni/src/main/java/libcore/io/Posix.java
@@ -916,7 +916,7 @@ public final class Posix implements Os {
         return -1;
     }
     int rc =
-      TEMP_FAILURE_RETRY(pwrite64([fd getInt$], bytes + bufferOffset, byteCount, offset));
+      TEMP_FAILURE_RETRY(pread64([fd getInt$], bytes + bufferOffset, byteCount, offset));
     return LibcoreIoPosix_throwIfMinusOneWithNSString_withInt_(@"pread", rc);
   ]-*/;
 


tomball:
Merged in change, will push publicly after code review.

tomball:
Added gist as unit test in new jre_emul/Tests/libcore/io/PosixTest.java source.

PR #557, opened 2015-06-25
Fix infix expression with extended operands

### LINKED ISSUE #556: Infix expression with extended operands like 1 << 2 << 3 not correctly translated
The expression `(1 << 2 << 3)` should be translated to `LShift32(LShift32(1, 2), 3)` but is currently translated to `LShift32(1, 2, 3)` in the [current master](https://github.com/google/j2objc/tree/51027f450381dfd75a4944bfb9219cea712f37f7) (version [0.9.7](https://github.com/google/j2objc/releases/tag/0.9.7) ignores the extended operands altogether).

lukhnos:
This fixes #556, that infix expressions with extended operands, such as `(1 << 2 << 3)`, are not correctly translated. In addition, two tests are added to `OperatorRewriterTest`.


tomball:
Thanks for the contribution!

PR #559, opened 2015-06-29
Fix pre/postincrement of parenthesized array access (fixes #558)

### LINKED ISSUE #558: Pre-/postincrement of parenthesized array access is mistranslated
If a pre- or postincrement expression involves a parenthesized array access, it is not correctly translated. Consider these expressions:

```java
// int x[] = { 0 };
(x[0])++;
++(x[0]);
```

The translated code currently looks like this:

```java
(IOSIntArray_Get(x, 0))++;
++(IOSIntArray_Get(x, 0));
```

Whereas the unparenthesized versions, `x[0]++` and `++x[0]`, are correctly translated:

```java
++(*IOSIntArray_GetRef(x, 0));
(*IOSIntArray_GetRef(x, 0))++;
```

lukhnos:
This fixes #558, that pre-/postincrement expressions containing parenthesized array accesses like `(x[0])` are mistranslated. Parenthesized expressions are now "unwrapped" first in `ArrayRewriter.needsAssignableAccess` so that the outermost prefix or postfix expression can be considered with the actual array access expression. Test cases are also added to `ArrayRewriterTest`.

One thing I noticed is that, after this fix, these two expressions will have the same translated code:

```java
x[0]++;     // (*IOSIntArray_GetRef(x, 0))++;
(x[0])++;   // (*IOSIntArray_GetRef(x, 0))++;
```

But:

```java
((x[0]))++; // ((*IOSIntArray_GetRef(x, 0)))++;
```

And this is because `ComplexExpressionExtractor` [parenthesizes an expression](https://github.com/google/j2objc/blob/f72b840c21ecf3ef8e09c36d370ba743acf77b22/translator/src/main/java/com/google/devtools/j2objc/translate/ComplexExpressionExtractor.java#L159) if its parent is a pre-/postfix expression, but does nothing if it is already a `ParenthesizedExpression`.

### DIFF
--- a/translator/src/main/java/com/google/devtools/j2objc/translate/ArrayRewriter.java
+++ b/translator/src/main/java/com/google/devtools/j2objc/translate/ArrayRewriter.java
@@ -25,6 +25,7 @@ import com.google.devtools.j2objc.ast.FunctionInvocation;
 import com.google.devtools.j2objc.ast.InstanceofExpression;
 import com.google.devtools.j2objc.ast.MethodInvocation;
 import com.google.devtools.j2objc.ast.NumberLiteral;
+import com.google.devtools.j2objc.ast.ParenthesizedExpression;
 import com.google.devtools.j2objc.ast.PostfixExpression;
 import com.google.devtools.j2objc.ast.PrefixExpression;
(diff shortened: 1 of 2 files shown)


kstanger:
Thanks for the fix!

PR #564, opened 2015-07-02
Use Class.isAssignableFrom to implement Class.asSubclass (fixes #563)

### LINKED ISSUE #563: Class.asSubclass does not work with interface types
Currently `-[IOSClass asSubclass]` always throws `JavaLangClassCastException` and only `IOSConcreteclass` implements the method, which in turn only works if the argument is also an `IOSConcreteclass`. This means you cannot do this:

```java
interface P {}
interface Q extends P {}
class A implements P {}

void f() {
    // Both valid in Java, but throw JavaLangClassCastException in Objective-C.
    A.class.asSubclass(P.class);
    Q.class.asSubclass(P.class);
}
```

Other `Class.asSubclass` implementations (like [this](http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/lang/Class.java#l3266) and [this](https://android.googlesource.com/platform/libcore/+/android-5.1.1_r6/libart/src/main/java/java/lang/Class.java)) are defined with `isAssignableFrom`, and this allows narrowing to and between interface types.

lukhnos:
This fixes #563 and bases `-[IOSClass asSubclass]` on `-[IOSClass isAssignableFrom]`. This takes away the need to implement `-asSubclass` in `IOSConcreteclass` and bring the behavior in line with other implementations ([example](http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/lang/Class.java#l3266)). A test case is also supplied.

### DIFF
--- a/jre_emul/Tests/com/google/j2objc/ClassTest.java
+++ b/jre_emul/Tests/com/google/j2objc/ClassTest.java
@@ -327,24 +327,24 @@ public class ClassTest extends TestCase {
 
   public void testAsSubclassCalls() throws Exception {
     assertTrue(canCallAsSubclass(Integer.class, Object.class));
-    assertTrue(!canCallAsSubclass(Long.class, Integer.class));
+    assertFalse(canCallAsSubclass(Long.class, Integer.class));
     assertTrue(canCallAsSubclass(Integer[].class, Object[].class));
     assertTrue(canCallAsSubclass(Integer[].class, Object.class));
-    assertTrue(!canCallAsSubclass(int.class, Object.class));
-    assertTrue(!canCallAsSubclass(int.class, long.class));
+    assertFalse(canCallAsSubclass(int.class, Object.class));
+    assertFalse(canCallAsSubclass(int.class, long.class));
     assertTrue(canCallAsSubclass(int[].class, Object.class));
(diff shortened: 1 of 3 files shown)


jre_emul/Tests/com/google/j2objc/ClassTest.java
```diff
@@ -311,6 +311,75 @@ public void testCharSequenceClass() throws Exception {
...
+    }
+
+    boolean assignable = y.isAssignableFrom(x);
+    assertEquals(assignable, callSuccessful);
+    return callSuccessful;
+  }
+
+  public void testAsSubclassCalls() throws Exception {
+    assertTrue(canCallAsSubclass(Integer.class, Object.class));
+    assertTrue(!canCallAsSubclass(Long.class, Integer.class));
```
kstanger: Please use "assertFalse()" instead of the "!" operator.

kstanger:
Thanks again for another fix. It looks good, just one comment to resolve.

lukhnos:
I updated the assertions. Let me know if you prefer the two commits be squashed into one or if you need anything else. Thanks for the review!

PR #570, opened 2015-07-10
Keep the do statement body if the const expression is false 

### LINKED ISSUE #569: The body of a do statement with constant false expression is erroneously skipped
This currently does not get translated to what you expect:

```java
// It's very unlikely that the Objective-C code returns 1.
int f() {
    int x;
    do {
        x = 1;
    } while (false);
    return x;
}
```

The translated code turns out to be:

```objc
- (jint)f {
  jint x;
  return x;
}
```

And this crashes j2objc:

```java
int g() {
    int x;
    foo:
        do {
            x = 1;
        } while(false);
    return x;
}
```

[`StatementGenerator.java:628`](https://github.com/google/j2objc/blob/a5b32a1fcfe6759bb5908ade5a4e6e94fbafcb26/translator/src/main/java/com/google/devtools/j2objc/gen/StatementGenerator.java#L628) throws `NullPointerException` when j2objc tries to generate the output. The crashing line wants to do `node.getBody().accept(this)`, but 
`node.getBody()` (`node` is an instance of `LabeledStatement`) returns `null`.

It appears that `ConstantBranchPruner.endVisit(DoStatement)` removes the `do` statement's body entirely even though it should be kept in the `while (false)` case.

lukhnos:
This fixes #569, that the body of a `do` statement with constant `false` expression is erroneously skipped. A test case is also added.

### DIFF
--- a/translator/src/main/java/com/google/devtools/j2objc/translate/ConstantBranchPruner.java
+++ b/translator/src/main/java/com/google/devtools/j2objc/translate/ConstantBranchPruner.java
@@ -42,7 +42,7 @@ public class ConstantBranchPruner extends TreeVisitor {
   @Override
   public void endVisit(DoStatement node) {
     if (getValue(node.getExpression()) == FALSE) {
-      node.remove();
+      node.replaceWith(node.getBody().copy());
     }
   }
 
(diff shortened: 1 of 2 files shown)


tomball:
As explained in [issue 570](https://github.com/google/j2objc/pull/570), the constant propagation optimizing works as intended. For the case where the do statement has a label and so j2objc fails with an NPE, an empty statement should be substituted.

tomball:
Thanks for the clarification in #569, reopening and will pull.

tomball:
Thanks for the bug report and fix!

PR #577, opened 2015-08-02
Create NewsApplication V1.06

aakkbb:


tomball:
There's no contribution here, just one new file that has a single empty line in it.

PR #591, opened 2015-08-26
Implement JreAssert with macro so that msg is conditionally evaluated

### LINKED ISSUE #590: The msg in JreAssert(cond, msg) should not be evaluated if cond is true
According to [14.10](https://docs.oracle.com/javase/specs/jls/se8/html/jls-14.html#jls-14.10) of the Java Language Specification, the `expression2` in the statement `assert expression1 : expression2` should only be evaluated when `expression1` is true. However, since `JreAssert(cond, msg)` is implemented as an inline function, `msg` will always be evaluated.

Here is a case, adapted from an open source project, where this can be a problem:

```java
assert x == null || x.size() == 0: "x must be null or empty, not of size " + x.size();
```

When `x` is `null`, the expression1 is evaluated to true. `x.size()` should never be evaluated, but currently this will crash the translated Objective-C program because `x.size()` in expression2 gets evaluated.

lukhnos:
This fixes #590 by using macro to implement `JreAssert`. A new test case is also added and has been run first without then with the change.

### DIFF
--- a/jre_emul/Classes/J2ObjC_source.h
+++ b/jre_emul/Classes/J2ObjC_source.h
@@ -44,13 +44,11 @@ __attribute__((always_inline)) inline id check_protocol_cast(
 
 FOUNDATION_EXPORT void JreThrowAssertionError(id __unsafe_unretained msg);
 
-__attribute__((always_inline)) inline void JreAssert(jboolean cond, id __unsafe_unretained msg) {
 #ifndef NS_BLOCK_ASSERTIONS
-  if (!cond) {
-    JreThrowAssertionError(msg);
-  }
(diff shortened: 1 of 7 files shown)


kstanger:
The problem with using a macro is that it will fail to compile if either the condition or message expression contain a comma. This was part of the reason that I added the inline function.

lukhnos:
I realized that's what `J2OBJ_COMMA` was for before 7d5d2877814b32bf33150490760487a56a52520f. Can't we bring part of it back?

kstanger:
Actually, if we just surround each expression with (), then the commas should be safe and we can use a macro. Can you add that change in StatementGenerator.java?

lukhnos:
> Can you add that change in StatementGenerator.java?

Sure thing!

lukhnos:
@kstanger This is ready for review. Thanks!

PR #595, opened 2015-08-27
Fix memory leaks in JavaMetadata

lukhnos:
This adds `autorelease` to three allocation sites in `JavaMetadata`. I found `-allFields` was leaking `JavaFieldMetadata` instances, and thought other two places also needed `autorelease`.


tomball:
Thanks for the fix!

Issue #599, opened 2015-09-02
Dead code elimination overkills/underkills in a few cases


lukhnos:
I found two types of cases for which the dead code eliminator currently can't generate compilable code. Here's the first type:

```java
class Foo {
  static final int x = f();

  static int f() { return 0; }

  static class Bar {}
}

class Baz extends Foo.Bar {}
```

The ProGuard usage report is just one line:

    Foo

So `Foo` is dead, but `Foo.Bar` and `Baz` aren't. In the code generated by `j2objc --dead-code-report`, however, `Foo.Bar` is not found, and `Foo.x` is kept. In the translated `+[Foo initialize]`, `Foo_x_` will get its value by calling `Foo_f()`, which no longer exists.

This is because `DeadCodeEliminator` removes all declarations inside a dead class except inlinable constants, and `Foo.x` is seen as such a constant because it's a static final primitive.

We should check if `Foo.x` is really a constant. As for static nested classes, even if both `Baz` and `Foo.Bar` are dead, `Foo.Bar` still shouldn't be removed, because of the way j2objc translates dead classes and interfaces. There will still be a `@interface Baz : Foo_Bar` generated, and therefore the removal of `Foo.Bar` is problematic.

The second type is about constructors:

```java
class P {
  String x;
  P(String x) {}
}

class Q extends P {
  Q(String x) { super(x); }
}

class R extends P {
  R(String x) { super(x); }
  static void f() {}
}
```

And the ProGuard usage report is:

    Q
    R:
        R(java.lang.String)

The generated `Q_init_` and `R_init_` read:

```objc
void Q_init(Q *self) {
  P_init(self);
}

void R_init(R *self) {
  P_init(self);
}
```

The problem is there is no `P_init`. The sole constructor of `P` requires a `String`, and so the generated init function is `P_initWithNSString_`.

`InitializationNormalizer.addDefaultConstructor` is called in both cases, but neither `Q_init` nor `R_init` has the information that `P`'s constructor requires. Since no instances will be created for `Q` and `R`, those generated default constructors will never be called, and I think it is sound to make them no-ops.

lukhnos:
PR merged, closing this now.

PR #600, opened 2015-09-02
Improve dead code elimination

lukhnos:
This fixes the cases discussed in #599 with changes to `DeadCodeEliminator`. This requires that the dead code map be passed to `InitializationNormalizer` later in the translation process, though. I've also added the discussed test cases. Improvements include:

* Don't remove static nested classes from a dead class
* Also visit static nested classes in a dead class
* Be more strict about what makes a value inlinable
* Don't call super from the generated default constructor if the entire class is dead or if all constructors are dead


tomball:
Thanks very much for your contribution!

PR #606, opened 2015-09-08
Fix the "lock = this" pattern in Reader, Writer, and synchronized Collections

### LINKED ISSUE #605: The translated java.io.Reader leaks memory when the lock object is itself
This is related to #169. `StringReader` is currently leaking itself, because the default constructor of `java.io.Reader`, its superclass, is using itself as the lock object:

``` java
protected Reader(Object lock) {
  if (lock == null) {
    throw new NullPointerException("lock == null");
  }
  this.lock = lock;
}
```

A [proposed fix](https://groups.google.com/d/msg/j2objc-discuss/fRgehuDZG5k/8MUFFzNfJP0J) discussed in the mailing list is to nullify `lock` in `StringReader.close`:

``` java
final StringReader reader = new StringReader(txt)  {
   public void close() {
    super.close();
    lock = null;
   }
};
```

This can be applied to `StringReader`'s own `close` method, but it is pretty common that `StringReader.close` is never called (presumably because people know that the underlying resource is a string).

I'm currently relying on modifying the translated `Reader.m`:

``` objc
- (void)dealloc {
  if (lock_ != self) {
    RELEASE_(lock_);
  }
  [super dealloc];
}

void JavaIoReader_init(JavaIoReader *self) {
  NSObject_init(self);
  // JreStrongAssign(&self->lock_, self);  // deleted
  self->lock = self;
}
```

Not sure if this is the best way to do it (and certainly modifying the translated output is not ideal), though it requires no changes to the `@synchorized(lock_)` uses in `Reader`'s other methods.

I haven't tried to move this back to the Java source using OCNI, though I assume it'll be needed (I realized, for example, that `IosX509Certificate` has its `-dealloc` implemented that way).

lukhnos:
This fixes #605 following @tomball's suggestion that we can just use a new object for the lock.

I'm a little curious what happens when j2objc wants to merge the next version of Android source. Would this kind of fix cause future maintenance issues?

### DIFF
--- a/jre_emul/android/libcore/luni/src/main/java/java/io/Reader.java
+++ b/jre_emul/android/libcore/luni/src/main/java/java/io/Reader.java
@@ -43,11 +43,12 @@ public abstract class Reader implements Readable, Closeable {
     protected Object lock;
 
     /**
-     * Constructs a new {@code Reader} with {@code this} as the object used to
-     * synchronize critical sections.
+     * Constructs a new {@code Reader} with a lock object to synchronize
+     * critical sections.
      */
     protected Reader() {
-        lock = this;
+        // j2objc: Use a new object rather than this to avoid reference cycle.
+        lock = new Object();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


tomball:
Would you please add a comment above each line, something like "// j2objc: use a new object rather than this to avoid reference cycle."? That will make it more obvious that this change needs to be preserved when merging with future Android releases. Answer your question? ;-)

lukhnos:
> That will make it more obvious that this change needs to be preserved when merging with future Android releases. Answer your question? ;-)

Makes sense. Comments added. Thanks!

Issue #622, opened 2015-09-30
ThreadPoolExecutor releases tasks prematurely when interrupted


danieldickison:
When `cancel(true)` is called on a Future that is currently being run by a ThreadPoolExecutor, the Callable object is released immediately which can lead to null pointer dereferencing if the body of the `call()` method accesses `this`.

While it's possible in some cases to catch an InterruptedException in the `call()` method, this isn't always possible (e.g. when doing networking with HttpURLConnection), and checking `Thread.interrupted()` is unreliable since the interrupt can happen between the check and subsequent code.

The bug is observed both with `-use-arc` and without. (Sorry about multiple edits here — initially I thought it only happened with -use-arc; it appears to happen less reliably without ARC, but it still does happen.)

Example:

```java
package com.bandcamp.shared;

import android.util.Log;

import java.util.LinkedList;
import java.util.List;
import java.util.concurrent.Callable;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.Future;

public class ExecutorServiceTest {
    private static final int THREADS = 2;
    private static final int DELAY = 1000;
    private String x = "something";
    private List<Future<Object>> futures = new LinkedList<>();
    private ExecutorService execServ = Executors.newFixedThreadPool(THREADS);

    public void test() {
        Future<Object> f = execServ.submit(new Callable<Object>() {
            @Override
            public Object call() throws Exception {
                doSomething();
                try {
                    Thread.sleep(DELAY);
                }
                catch (InterruptedException e) {}
                Thread.sleep(DELAY);
                doSomething(); // Crashes here
                return null;
            }
        });
        try {
            Thread.sleep(DELAY / 2);
            f.cancel(true);
        } catch (InterruptedException e) {
            e.printStackTrace();
        }

        //futures.add(f); // prevents crash
    }

    private void doSomething() {
        Log.i("ExecutorServiceTest", "doSomething: " + x);
    }
}
```

```swift
let executorServiceTest = BCExecutorServiceTest()
executorServiceTest.test()
```

danieldickison:
One workaround is to enable `-use-arc` and then create a strong reference to `this` for the duration of `call()`:

```java
public Object call() throws Exception {
    Object strongThis = this;
    strongThis.hashCode(); // quick way to suppress unused var warnings in both java and objc.
    ...
```

kstanger:
An alternative to `-use-arc` in your workaround is our `@RetainedLocalRef` annotation:
[...]

PR #623, opened 2015-10-01
Fix premature release in canceled FutureTask

danieldickison:
If a FutureTask is already running when it is canceled by another thread, the Callable in the FutureTask is released immediately and the Callable.call method can crash it it references `this`. This fix makes sure the FutureTask keeps a local retained reference to the Callable while it runs.

Fixes issue #622 (ThreadPoolExecutor releases tasks prematurely when interrupted).

### DIFF
--- a/jre_emul/android/libcore/luni/src/main/java/java/util/concurrent/FutureTask.java
+++ b/jre_emul/android/libcore/luni/src/main/java/java/util/concurrent/FutureTask.java
@@ -274,7 +274,7 @@ public class FutureTask<V> implements RunnableFuture<V> {
         boolean ran = false;
         int s = state;
         try {
-            Callable<V> c = callable;
+            @RetainedLocalRef Callable<V> c = callable;
             if (c != null && s == NEW) {
                 try {
                     c.call(); // don't set result
--- a/jre_emul/android/libcore/luni/src/main/java/java/util/concurrent/FutureTask.java
+++ b/jre_emul/android/libcore/luni/src/main/java/java/util/concurrent/FutureTask.java
@@ -6,6 +6,8 @@
 
 package java.util.concurrent;
 
+import com.google.j2objc.annotations.RetainedLocalRef;
+
 import java.util.concurrent.locks.LockSupport;
 
 /**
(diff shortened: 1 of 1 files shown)


jre_emul/android/libcore/luni/src/main/java/java/util/concurrent/FutureTask.java
```diff
@@ -229,7 +231,7 @@ public void run() {
             !U.compareAndSwapObject(this, RUNNER, null, Thread.currentThread()))
             return;
         try {
-            Callable<V> c = callable;
+            @RetainedLocalRef Callable<V> c = callable;
```
kstanger: Please duplicate this change below. (line 277)

kstanger:
Great job, I also tracked this issue down and have the same change locally!

I should point out that even with this fix, there still exists a race condition because the instructions that read the "callable" field then retain it do not do so atomically. I'm currently looking into modifications to our memory model to help with concurrent field access. In the meantime, this should fix your issue from a practical point of view.

jre_emul/android/libcore/luni/src/main/java/java/util/concurrent/FutureTask.java
danieldickison: Good catch; done.

danieldickison:
Thanks! I did notice the non-atomicity of reading `callable` and then retaining it, but hopefully the window for this race condition is small enough to be a rare condition in practice. It seems difficult to overcome without some sort of locking in the FutureTask implementation (making atomic the variable read+retain when running and the write+release when canceling). I'm curious to see what changes to the memory model will solve this issue.

kstanger:
It seems that solving the race condition will require locking on variable access, since there's no other way to make the read+retain atomic. We'll probably want to limit this to volatile variables, in which case the fix here will be to declare callable as volatile.

danieldickison:
Thanks for merging. I agree that a lock would be necessary. I thought maybe ARC can guarantee that atomicity without locks, but I don't think that's the case after reading this section of the ARC documentation: http://clang.llvm.org/docs/AutomaticReferenceCounting.html#optimization

kstanger:
Also, here is Apple's implementation for synthesized properties, which uses
spinlocks for atomic properties.
http://www.opensource.apple.com/source/objc4/objc4-532.2/runtime/Accessors.subproj/objc-accessors.mm

kstanger:
FYI - I've documented the recent changes to the memory model here: http://j2objc.org/docs/Memory-Model.html

danieldickison:
That's super useful and also fascinating to read. Thanks Keith.

PR #638, opened 2015-10-15
Swift frames in stack traces

danieldickison:
Swift frames are currently excluded from stack traces (e.g. when logging exceptions and via `Thread.getStackTrace()`) because Swift mangles symbol names. This patch adds native code in StackTraceElement.java to attempt to demangle the most common forms of Swift function symbols so that they appear in j2obc-generated stack traces.

As far as I know, the name mangling scheme is undocumented. This implementation uses info from several sources online where people have reverse-engineered the format, as well as some of my own experimentation. It doesn't handle every possible Swift symbol, but it should handle most "normal" object methods, and it should bail out safely if it encounters anything that it doesn't recognize.

It would be nice to have tests for this, but I haven't figured out how to incorporate Swift-based unit tests.

### DIFF
--- a/jre_emul/Classes/java/lang/StackTraceElement.java
+++ b/jre_emul/Classes/java/lang/StackTraceElement.java
@@ -168,15 +168,15 @@ public class StackTraceElement implements Serializable {
 
   /*-[
   // This is based on undocumented details of the Swift compiler as reverse-engineered by several
-  // sources online, and it doesn't attempt to be comprehensive, but tries to demangle the most common
-  // type of object instance methods.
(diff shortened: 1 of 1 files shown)


tomball:
Thanks, this looks great! (And yes, we need to figure out Swift unit testing soon.)

Please reformat for a maximum of 100 columns, as otherwise our internal version control system balks. It looks like that only affects one statement and two comment lines.

danieldickison:
@tomball: done.

I'm sure more and more new projects will start using Swift on the iOS side, so having unit tests that run Swift will be a big benefit.

PR #642, opened 2015-10-16
Update Number.java for Long serializable

MarcImbert:
Hi,
Sorry in advance for my poor english.
I've an error when I translate a class at who I pass a "Long" type when it expect a generic type serializable and comparable:
error: SimpleAutoIndexedEntity.java:15: Bound mismatch: The type Long is not a valid substitute for the bounded parameter <ID extends Serializable & Comparable<ID>> of the type SimpleEntity<ID>
It seems that it was because the Number.java (who is present in the jre_emul.jar file) doesn't implement java.io.Serializable. That's why "Long" class who inherit of the "Number" class can't be serializable.
I've compared to the official Number.java file from the java.lang package and there is other differences on byteValue and shortValue methods.
I don't know if it's necessary, or obligatory, to update this methods too, but this update seems work with it.
Do you think it's good?

### DIFF
--- a/jre_emul/stub_classes/java/lang/Number.java
+++ b/jre_emul/stub_classes/java/lang/Number.java
@@ -22,12 +22,12 @@ package java.lang;
  *
  * @see java.lang.Object
  */
-public abstract class Number {
+public abstract class Number implements java.io.Serializable {
 
   public Number() {}
 
   public byte byteValue() {
-	  return 0;
+	  return (byte)intValue();
   }
 
   public abstract double doubleValue();
@@ -39,6 +39,8 @@ public abstract class Number {
   public abstract long longValue();
 
   public short shortValue() {
-    return 0;
+    return (short)intValue();
   }
+  
+  private static final long serialVersionUID = -8742448824652078965L
 }
--- a/jre_emul/stub_classes/java/lang/Number.java
+++ b/jre_emul/stub_classes/java/lang/Number.java
@@ -27,7 +27,7 @@ public abstract class Number implements java.io.Serializable {
   public Number() {}
 
   public byte byteValue() {
-	  return (byte)intValue();
+	  return 0;
   }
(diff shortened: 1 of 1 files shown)


MarcImbert:
I signed it.

De : googlebot [mailto:notifications@github.com]
Envoyé : vendredi 16 octobre 2015 12:42
À : google/j2objc
Cc : marc imbert
Objet : Re: [j2objc] Update Number.java for Long serializable (#642)


Thanks for your pull request. It looks like this may be your first contribution to a Google open source project, in which case you'll need to sign a Contributor License Agreement (CLA).

[:memo:]Please visit https://cla.developers.google.com/ to sign.

Once you've signed, please reply here (e.g. I signed it!) and we'll verify. Thanks.

________________________________

  *   If you've already signed a CLA, it's possible we don't have your GitHub username or you're using a different email address. Check your existing CLA data<https://cla.developers.google.com/clas> and verify that your email is set on your git commits<https://help.github.com/articles/setting-your-email-in-git/>.
  *   If you signed the CLA as a corporation, please let us know the company's name.

tomball:
Thanks for investigating this issue. Number is mapped directly to NSNumber, rather than being translated, so this is a stub class which is only used for its class and method declarations. That means that only the 'implements Serializable' part of your fix helps translation. The translator maps [NSNumber's charValue to Number.byteValue, and longLongValue to Number.longValue](https://github.com/google/j2objc/blob/master/translator/src/main/resources/com/google/devtools/j2objc/JRE.mappings#L171). None of the methods in stub classes do anything, so we make them empty or return false/0/null to make that obvious.

Please undo the method changes and I'll be happy to pull in your 'implements Serializable' fix.

MarcImbert:
I've undo the method changes.

tomball:
Thanks for your contribution!

PR #649, opened 2015-10-26
Segmented-headers defaults to true

advayDev1:
Adds the flag --no-segmented-headers to disable the feature.
Ignore and accept the old --segmented-headers flag.

Per https://groups.google.com/d/msg/j2objc-discuss/Db3YUD9V0TY/C4HYerpQAwAJ

### DIFF
--- a/translator/src/test/java/com/google/devtools/j2objc/gen/ObjectiveCSegmentedHeaderGeneratorTest.java
+++ b/translator/src/test/java/com/google/devtools/j2objc/gen/ObjectiveCSegmentedHeaderGeneratorTest.java
@@ -30,7 +30,7 @@ public class ObjectiveCSegmentedHeaderGeneratorTest extends GenerationTest {
   @Override
   protected void setUp() throws IOException {
     super.setUp();
-    Options.enableSegmentedHeaders();
+    // Segmented headers are on by default.
   }
 
   public void testTypicalPreprocessorStatements() throws IOException {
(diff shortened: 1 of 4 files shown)


advayDev1:
passes on Travis: https://travis-ci.org/madvay/j2objc/jobs/87348237
not sure what else i should test.

ptal @kstanger

tomball:
Advay, is there a way you can strip down the include file problem that prevented your using the default behavior. We may not be able to fix it (C will always have the potential for include file cycles), but it helps knowing the pattern.

advayDev1:
I'm not having any specific problem actually.  The reason I asked about --segmented-headers being a default on the mailing list was that various open-source libraries need the option to build successfully.

If the option doesn't break anything, and for some cases it fixes things, it seems like it should be the default (even if it doesn't fix all possible include cycles).

translator/src/test/java/com/google/devtools/j2objc/gen/ObjectiveCSegmentedHeaderGeneratorTest.java
```diff
@@ -30,7 +30,7 @@
   @Override
   protected void setUp() throws IOException {
```
kstanger: Please remove the setUp() method entirely.
Also, please remove the declaration of Options.enableSegmentedHeaders() since it is no longer used.

advayDev1: done

kstanger:
Sorry this took so long. It's been a slow week for me.

PR #655, opened 2015-10-29
Eliminate inner class constructors correctly when ProGuard explicitly instructs so

### LINKED ISSUE #654: Inner class constructor not removed when ProGuard report explicitly lists it
Here is an edge case of the dead code eliminator. Usually if an inner class is not used, the class will just be a one-line entry in the ProGuard report:

```
Outer$Inner
```

But if for some reason ProGuard decides to keep the class but to remove all its fields (such as when some `-keep class com.example.*` is used), it lists all the fields in the inner class instead. For example:

```
Outer$Inner:
    int z
    final synthetic Outer this$0
    9:9:Outer$Inner(Outer,int)
    10:10:int f()
```

The problem is that J2ObjC does not remove the inner class constructor as instructed. Here's the Java code used:

```java
public class Outer {
  int y;
  public Outer(int x) { y = x; }

  class Inner {
    int z;
    Inner(int x) { z = x; }
    int f() { return z + y; }
  }
}
```

I've also made a [sample project](https://github.com/lukhnos/dead-code-elim-edge-case-j2objc) to demo the bug.

The cause is that dead code elimination is the first pass in the translation processor, and therefore `BindingUtil.getSignature` cannot return well-formed inner class constructor signature, since those methods are not yet furnished with outer types by a later pass (I believe it's `InnerClassExtractor`).

I have an upcoming patch that adds a new `getProGuardSignature` method to `BindingUtil` though I wonder if the dead code elimination pass can be run at a later stage.

lukhnos:
This fixes #654 by adding a new `getProGuardSignature` method to `BindingUtil` so that the dead code eliminator can use the correct signature to match against the dead code report. As mentioned in #654, the signature was not correct because dead code elimination is the first pass in the translation processor, before inner class methods are furnished with outer types in a later pass.

Since `InitializationNormalizer` needs the dead code map to work correctly, I wonder if dead code elimination should be done after `InnerClassExtractor` is finished, but I'm not sure if the pass will still work after the nodes have been modified by the other passes. The downside of the approach here is one more special case in `BindingUtil`, though I believe it's straightforward enough for the job.

This also fixes [a previous test case](https://github.com/lukhnos/j2objc/commit/0024184c7ea56c399149974c2f732b5a2641557f), as I don't think ProGuard produces the same signature.

### DIFF
--- a/translator/src/main/java/com/google/devtools/j2objc/translate/DeadCodeEliminator.java
+++ b/translator/src/main/java/com/google/devtools/j2objc/translate/DeadCodeEliminator.java
@@ -155,7 +155,7 @@ public class DeadCodeEliminator extends TreeVisitor {
         }
         IMethodBinding binding = method.getMethodBinding();
         String name = getProGuardName(binding);
-        String signature = BindingUtil.getSignature(binding);
+        String signature = BindingUtil.getProGuardSignature(binding);
         if (deadCodeMap.isDeadMethod(clazz, name, signature)) {
           if (method.isConstructor()) {
             deadCodeMap.addConstructorRemovedClass(clazz);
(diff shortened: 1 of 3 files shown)


tomball:
Thanks for the fix! I'm also nervous about rearranging the translation phases, as there are several dependencies (just like in a compiler).

PR #670, opened 2015-12-02
Rewrote link flags doc.

tomball:
You can see the generated page [here](http://tomball.github.io/j2objc/docs/Required-Link-Flags.html) (ignore the screwed up menu on the left).

### DIFF
--- a/docs/Required-Link-Flags.md
+++ b/docs/Required-Link-Flags.md
@@ -14,67 +14,45 @@ the [j2objcc](j2objcc.html) command-line script, but need to be specified when b
 J2ObjC's distribution includes several static libraries; to use them, your project needs to tell
 the linker where to find them.
 
-Generally, the library search path needs to include "$(j2objc_distribution)/lib", where the
-"$(j2objc_distribution)" variable is the path to your local copy of J2ObjC.  For example, if you
+Generally, the library search path needs to include _$(j2objc_distribution)/lib_, where the
+_$(j2objc_distribution)_ variable is the path to your local copy of J2ObjC.  For example, if you
 unzipped a [J2ObjC release archive file](https://github.com/google/j2objc/releases) to 
 "/usr/local/", this path would be "/usr/local/j2objc".
 
-**Important**: Do not actually use "$(j2objc_distribution)" in your project; always specify the
+**Important**: Do not actually use _$(j2objc_distribution)_ in your project; always specify the
 actual path where you installed J2ObjC.
 
-If you build J2ObjC from a copy of its source code, then "$(j2objc_distribution)" is your copy's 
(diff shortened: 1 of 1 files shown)


docs/Required-Link-Flags.md
```diff
@@ -1,38 +1,172 @@
...
+the real path).
 
-### **-ObjC**
+## JRE Libraries
 
-This flag tells the linker to include all Objective-C categories in your application. J2ObjC requires this flag because it has categories that extend NSObject and NSString to be support `java.lang.Object` and `java.lang.String` functionality.
+These libraries implement classes defined by J2ObjC's JRE emulation:
 
-**Note**: if you get "unrecognized selector" link errors, this flag wasn't set.
+<table>
```
kstanger: Markdown provides table syntax, which I use in Translation-Reference.md (https://raw.githubusercontent.com/google/j2objc/gh-pages/docs/Translation-Reference.md). It looks like it will generate the same HTML, but is easier to read in text.

docs/Required-Link-Flags.md
```diff
@@ -1,38 +1,172 @@
...
+    </td>
+  </tr><tr>
+    <td>libjre_io.a</td>
+    <td style="white-space: nowrap;">-l jre_io</td>
+    <td>
+      Several classes in the `java.io` package. Many `java.io` classes are in `jre_core`, though,
+      so only include this library if there are unresolved JavaIo* symbol errors.
+    </td>
+  </tr>
+</table>
```
kstanger: This is missing a few of the libraries. Do you plan to add them in later?

docs/Required-Link-Flags.md
```diff
@@ -1,38 +1,172 @@
...
-## JUnit Linker Flag
+### libjre_emul.a (**-l jre_emul**)
 
-When linking translated JUnit tests, add the following flag to the above:
-````
--l junit
-````
+The `jre_emul` library contains all the classes included in J2ObjC's JRE emulation. If an app is
+linked with `jre_emul`, none of the other jre_* libraries should be included, or the linker will
+report duplicate symbol errors. That is because `jre_emul` is includes all classes defined in
```
kstanger: "is includes" -> "includes"

docs/Required-Link-Flags.md
tomball: Turns out I couldn't use it after all, as word wrapping caused the "-l channels" flag to be on two lines. It also strips markdown and non-link HTML from the cells.

docs/Required-Link-Flags.md
tomball: Added.

docs/Required-Link-Flags.md
tomball: Fixed.

Issue #666, opened 2015-11-23
make dist fails


tm-linkwerk:
I just checked out j2objc. When I run make dist, compiling CheckForNull.m fails because

<pre>
error: unknown warning option '-Wno-nullability-completenes'; did you mean '-Wno-delete-incomplete'?
</pre>

When I run make -k dist, the same message is printed for many other files and at the end it says:

<pre>
make[1]: Target 'dist' not remade because of errors.
make: *** [jsr305_dist] Error 2
</pre>

OS X 10.9.5

tomball:
I'm able to build from j2objc's latest source without issue. 

One red flag in your report is that there is no reference to '-Wno-nullability-completenes' in our project -- all references use '-Wno-nullability-completeness' (two Ss at the end). Did you specify this flag anywhere in your project? It looks like a misspelling in your local build environment. Try running `git status` in your j2objc repository to see if some file was accidentally modified.

FWIW, a great text search utility I use all the time is [ack](http://beyondgrep.com/). It's like a combination of grep and find, but easier to use. Here's what I see running ack in my unmodified client:

```
tball:j2objc (master)$ ack Wno-nullability-completenes
guava/Makefile
36:  -Wno-nullability-completeness $(INCLUDE_DIRS:%=-I%)

jsr305/environment.mk
33:WARNINGS = -Wall -Werror -Wno-nullability-completeness
```

If it reports anything else, there's your problem.

tm-linkwerk:
I'm sorry, it was a misspelling when pasting into this issue, the correct message is

<pre>
error: unknown warning option '-Wno-nullability-completeness'; did you mean
      '-Wno-delete-incomplete'? [-Werror,-Wunknown-warning-option]
</pre>

Meanwhile I pulled again from the git repo, updated to the latest OS X 10.11.1 and Xcode 7.1.1 but the error persists.

tm-linkwerk:
I just discovered that there was an old 6.0.1 version of xcode installed beside the newer 7.1.1. So I removed it and I'm just running ```make dist``` again to remove any doubt.

tm-linkwerk:
Ok, building with Xcode 7.1.1 works.

Maybe README (https://github.com/google/j2objc/blob/master/README.md) and INSTALL (https://github.com/google/j2objc/blob/master/INSTALL) should be updated.

PR #672, opened 2015-12-07
Update build requirements and tool links

lukhnos:
Following the suggestions in #666, this updates the build requirements as well as links to the mentioned tools.


lukhnos:
The CI build failed as expected. It'll require #673 to get it back to normal.

PR #673, opened 2015-12-07
Upgrade Travis CI settings

lukhnos:
The Travis CI jobs of my fork was reporting the following errors: [this job](https://travis-ci.org/lukhnos/j2objc/jobs/95389403) failed because Travis CI's default OS X image was still on Xcode 6.1, then after upgrading the setting [this job](https://travis-ci.org/lukhnos/j2objc/jobs/95399149) failed because it appears that `brew-cask` has now been part of the recent OS X images.

After those changes it now builds, but [this job](https://travis-ci.org/lukhnos/j2objc/jobs/95404176) didn't finish because it exceeded the maximum 50 minutes that Travis allows. Perhaps we could disable `make dist` for now and only run the tests?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,8 @@ before_script:
   - scripts/osx-set-java.sh
 
 # Travis-CI kills the build after 50 minutes, so break it up into
-# smaller commands.
-script: make -j4 dist
+# smaller commands and only bulid two archs.
+script: J2OBJC_ARCHS="macosx iphone" make -j4 dist
 after_script:
   - make test_translator
   - make test_cycle_finder
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 # j2objc is a mixed-language project, but setting objective-c forces the OS to OSX.
 language: objective-c
 os: osx
+osx_image: xcode7.1
 
 env:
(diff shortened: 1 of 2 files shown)


tomball:
You mean just run the translator and maybe the cycle_finder tests? We could do that, but I'd like to avoid dropping the native build if we can avoid it.

Another way to cut down the build time is to only build one arch, maybe just macosx? That will verify that everything builds and JUnit tests can be run, though it won't catch any iOS-specific issues.

lukhnos:
@tomball Made a change in 57d9480d59bd7d5b9ba585ae2325391b88c5dc5a so that it builds `macosx` as well as `iphone`. I was hoping to also build `iphone64` so as to have a representative 64-bit iOS build, but that turned out to be one too many – the job still ran out of time.

My hope now is that the 64-bit OS X build suffices as a canary for any potential 64-bit-platform issues, and all other iOS-specific issues will be caught by the `iphone` build.

[The CI job for my branch](https://travis-ci.org/lukhnos/j2objc/builds/95440323) now finished in 36 minutes. I'll wait to see if [this PR's CI job](https://travis-ci.org/google/j2objc/builds/95440326) also finishes within that much time.

PR #676, opened 2015-12-09
Fix cyclic reference leaks in ReferenceQueue

### LINKED ISSUE #675: WeakHashMap leaks through ReferenceQueue
I found my app was leaking `JavaUtilWeakHashMap_Entry`. Digging further, I found the app was also leaking `JavaLangRefWeakReference`. The root cause appears to be that `ReferenceQueue` uses a cyclic reference when the queue only has one entry ([source](https://github.com/google/j2objc/blob/21a9ef27754da9a31bba262854cf74fbe94ccb87/jre_emul/android/libcore/luni/src/main/java/java/lang/ref/ReferenceQueue.java#L137)):

```java
reference.queueNext = reference;
```

As a result, this method leaks:

```java
@AutoreleasePool
public static void testWeakHashMap() {
  WeakHashMap<Object, String> m = new WeakHashMap<>();
  m.put(new Object(), "some value");
}
```

And this too (you'll recognize this is adopted from [WeakReferenceTest](https://github.com/google/j2objc/blob/21a9ef27754da9a31bba262854cf74fbe94ccb87/jre_emul/Tests/java/lang/ref/WeakReferenceTest.java#L79)):

```java
@AutoreleasePool
public static void testQueuedWeakReference() {
  final boolean[] dealloced = { false };
  ReferenceQueue<? super Object> queue = new ReferenceQueue<Object>();
  for (@AutoreleasePool int i = 0; i < 1; i++) {
    Object referent = new Object() {
      public void finalize() {
        dealloced[0] = true;
      }
    };
    weakRef = new WeakReference<Object>(referent, queue);
  }

  if (weakRef.get() != null) {
    throw new AssertionError("Should be null");
  }
}
```

I've made a [gist](https://gist.github.com/lukhnos/f628a45939cba16d3f11) that can be run in Instruments. A PR will also be up soon.

lukhnos:
This fixes #675 by using a sentinel object to mark that a `Reference` is the only entry in the queue. I followed @kstanger's [previous fix for ConcurrentLinkedQueue](https://github.com/google/j2objc/commit/db2597b64e3cd8ce18a6d8832df08b35a7a7b529) for the naming convention.

After this I no longer see the leaks in [my test case](https://gist.github.com/lukhnos/f628a45939cba16d3f11), and an app of mine that uses a lot of `WeakHashMap` no longer leaks.

One thing I don't fully understand is under which condition a `ReferenceQueue` may still have entries not removed by `poll()` (and therefore it leaves to the translated queue's `-dealloc` to clean up the `head`) , but I'm definitely interested in using this chance to learn more about J2ObjC's weak-referencing mechanism.

### DIFF
--- a/jre_emul/android/libcore/luni/src/main/java/java/lang/ref/ReferenceQueue.java
+++ b/jre_emul/android/libcore/luni/src/main/java/java/lang/ref/ReferenceQueue.java
@@ -29,6 +29,9 @@ public class ReferenceQueue<T> {
 
     private Reference<? extends T> head;
 
+    // j2objc: Use a sentinel to avoid the "reference.queueNext = reference" leak.
+    private final Reference<? extends T> SENTINEL = new WeakReference<>(null, null);
+
     /**
      * Constructs a new instance of this class.
      */
@@ -52,7 +55,7 @@ public class ReferenceQueue<T> {
 
         ret = head;
 
-        if (head == head.queueNext) {
+        if (SENTINEL == head.queueNext) {
             head = null;
         } else {
             head = head.queueNext;
@@ -134,7 +137,7 @@ public class ReferenceQueue<T> {
      */
     synchronized void enqueue(Reference<? extends T> reference) {
         if (head == null) {
-            reference.queueNext = reference;
+            reference.queueNext = SENTINEL;
         } else {
             reference.queueNext = head;
         }


kstanger:
This looks good. I'll wait for the Travis build result before merging.

PR #700, opened 2016-02-03
fixing head to build with public xmlpull headers

twoboxen:
XMLPull headers should be public (actually, most of these private headers should be).  I'm not sure of the advantage of hiding headers to statically linked classes.  Making them private just ensures that no one can use them (not even linking in their own transpiled then compiled versions).

Also, I had to take the translated headers copying out of a subshell to get "make dist" to work.  It seems that there is some sort of race condition.

### DIFF
--- a/jre_emul/Makefile
+++ b/jre_emul/Makefile
@@ -206,8 +206,8 @@ dist_includes: $(BUILD_DIR)/.jre_dist_includes
 $(BUILD_DIR)/.jre_dist_includes: $(PUBLIC_TRANSLATED_HEADERS:%=$(GEN_OBJC_DIR)/%) \
     $(PUBLIC_EMULATION_HEADERS:%=$(EMULATION_CLASS_DIR)/%) | $(DIST_INCLUDE_DIR)
 	@echo copying jre_emul header files
-	@(cd $(EMULATION_CLASS_DIR) && tar cf - \
-	    $(PUBLIC_EMULATION_HEADERS:$(EMULATION_CLASS_DIR)/%=%)) \
+	@cd $(EMULATION_CLASS_DIR) && tar cf - \
+	    $(PUBLIC_EMULATION_HEADERS:$(EMULATION_CLASS_DIR)/%=%) \
 	    | (cd $(DIST_INCLUDE_DIR); tar xfp -)
 	@(cd $(GEN_OBJC_DIR) && tar cf - \
 	    $(PUBLIC_TRANSLATED_HEADERS:$(GEN_OBJC_DIR)/%=%)) \
--- a/jre_emul/java_sources.mk
+++ b/jre_emul/java_sources.mk
@@ -1110,7 +1110,12 @@ JAVA_PUBLIC_SOURCES_XML = \
   org/xml/sax/helpers/ParserFactory.java \
   org/xml/sax/helpers/XMLFilterImpl.java \
   org/xml/sax/helpers/XMLReaderAdapter.java \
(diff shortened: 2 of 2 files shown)


twoboxen:
I signed it!

jre_emul/Makefile
```diff
@@ -206,8 +206,8 @@ dist_includes: $(BUILD_DIR)/.jre_dist_includes
 $(BUILD_DIR)/.jre_dist_includes: $(PUBLIC_TRANSLATED_HEADERS:%=$(GEN_OBJC_DIR)/%) \
     $(PUBLIC_EMULATION_HEADERS:%=$(EMULATION_CLASS_DIR)/%) | $(DIST_INCLUDE_DIR)
 	@echo copying jre_emul header files
-	@(cd $(EMULATION_CLASS_DIR) && tar cf - \
-	    $(PUBLIC_EMULATION_HEADERS:$(EMULATION_CLASS_DIR)/%=%)) \
+	@cd $(EMULATION_CLASS_DIR) && tar cf - \
```
kstanger: So this was the only header copying command that gave you issues? How strange.

At first glance it seems that removing the sub shell would leave the shell in a different directory (EMULATION_CLASS_DIR) when entering the sub shell on line 211. I believe this would have been the original reason for the sub shell. Does this not affect the final destination of the headers?

twoboxen: Yeah i agree.  all the files look ok, but I ran it a bunch of times until this finally worked.  I'm not sure that these lines should remain.  i can just hack these in whenever I need to create a build.  Though I have no idea why I'm the unique snowflake environment....

kstanger: It's possible that the subshell isn't necessary. According to this a subshell is implied by the pipe:
http://stackoverflow.com/questions/5760640/left-side-of-pipe-is-the-subshell

twoboxen: Well, that would certainly make it easier to read the makefiles... but
that's in a lot of places :-/

kstanger: When I try this change locally, I'm getting "./dist/include: No such file or directory". The pipe has higher precedence than the "&&" so the initial change directory is not run in a subshell.

Could you try switching the "&&" with a ";" while preserving the subshell brackets. Just to see if it works for you. Another thing to try is pushd and popd commands.

tomball: What shell are you running? (run "ps" to double-check) We all use the OSX
default, bash.

twoboxen: zsh, but i've tried bash, sh also (with same results)

twoboxen: Ok, I don't think that fix actually fixed anything (same problem)
copying jre_emul header files
tar: Unrecognized archive format
tar: Error exit delayed from previous errors.
make[1]: *** [/Users/mhawkins/src/j2objc/jre_emul/build_result/.jre_dist_includes] Error 1
make[1]: *** Waiting for unfinished jobs....
make: *** [jre_emul_dist] Error 2

I'm guessing the files were copied there when I ran the command manually... which allowed the build to finish.

If I ONLY change that line, the files aren't copied, but the "copy headers" phase completes.  The problem later comes when 'IOSClass.h' isn't found... because it wasn't actually copied to dist/include

twoboxen:
So I can't use HEAD due to https://github.com/google/j2objc/issues/701, but I found a fix to the problem you can see in this replacement PR (https://github.com/google/j2objc/pull/702).

This actually is a PR against the 1.0.0 tag.  we can close this one.

PR #731, opened 2016-03-24
Fixes a number of retain cycles and exceptions

gistya:
I have fixed the leaks. Most were from retain cycles, the rest from exceptions.
## Explanation of the retain cycles leaks

TransformerImpl - m_outputTarget is passed as an argument into a function and is owned by the caller, yet was being assigned to a strong instance variable of the callee. I set it to weak.

TransformerIdentityImpl - m_serializer gets set to a property of a property of a property of an argument passed to the caller from the callee. Should not be owned by callee. I set it to weak.

ElemContext - m_next is pretty clearly a cyclical reference, as Keith already noted in an e-mail. I set it to weak.

NamespaceMappings - m_namespaces and m_nodeStack are both collections that will end up containing things that are already owned elsewhere. I would think that resetting the Transformer should clear these, but I tried it, and that did not free up the memory or fix the leak. So, I set these to weak, and it fixed the leak without seeming to have any negative effects. 
## Examples of leaks from Exceptions

( 1 )

For example in the translated Encodings.m: 

```
@try {
NSString *highVal = [((NSString *) nil_chk([((NSString *) nil_chk(val)) substring:len])) trim];
highChar = (jchar) [((JavaLangInteger *) nil_chk(JavaLangInteger_decodeWithNSString_(highVal))) intValue];
}
@catch ... (throw an exception and set highChar = 0)
```

If highVal comes in as an empty string (this happens often), then an exception gets thrown. Memory leak. 

So I had to go in by hand and change it to:

```
         NSString *highVal = [((NSString *) nil_chk([((NSString *) nil_chk(val)) substring:len])) trim];
         if(highVal == nil) {
            highChar = 0;
          }
          else if([highVal isEqualToString:@""]) {
            highChar = 0;
          }
          else {
              @try {
                   highChar = (jchar) [((JavaLangInteger *) nil_chk(JavaLangInteger_decodeWithNSString_(highVal))) intValue];
               }
              @catch ... (throw an exception and set highChar = 0)
          }
```

**Solution**: Now that I know that works I have modified Java source to affect the same result (check for a null or empty string first before getting draconian with exceptions). 

( 2 )

In another case, I was seeing this call tree leading to an exception being thrown:

OrgApacheXmlSerializerCharInfo_initWithNSString_withNSString_withBoolean_ ---->
JavaUtilResourceBundle_getBundleWithNSString_withJavaUtilLocale_withJavaLangClassLoader_   ---->
JavaUtilResourceBundle_handleGetBundleWithBoolean_withNSString_withJavaUtilLocale_withJavaLangClassLoader_ ---->
JavaUtilResourceBundle_handleGetBundleWithBoolean_withNSString_withJavaUtilLocale_withJavaLangClassLoader_ ---->
IOSClass_forName_ ---->
[JavaLangClassNotFoundException initWithNSString:]

So I put in some debug code into JavaUtilResourceBundle to NSLog the name of the class that it was not finding. Here are the three classes it could not find:

org.apache.xml.serializer.XMLEntities_en_US
org.apache.xml.serializer.XMLEntities_en
org.apache.xml.serializer.XMLEntities

**Explanation**: 
These are not actually classes. Xalan's apache.xml.serializer.CharInfo (l. 194) actually calls getBundle on java.util.PropertyResourceBundle. But PropertyResourceBundle does not implement that method, so we use its super-class, ResourceBundle's static methods to try to find a resource called "XMLEntities.properties". 

However, when ResourceBundle's handleGetBundle runs, the first thing it tries to do is append a locale string to the end, then try to load that as a class. That throws an exception of course. Then it strips off "_US" and tries to load the remaining string again as a class. That throws an exception as well. Then it strips off "_en" and tries to load that as a class. That also throws an exception.

Only after throwing these there exceptions does it finally get around to adding ".properties" onto the end, and trying to load that.

Why are we even trying to load it as a class here? We shouldn't be; we knew it was a .properties file from the beginning. After all, the only really good reason to subclass ResourceBundle as PropertyResourceBundle would be precisely to not do try to load it as a class first since we already know it's a .properties file. 

**Solution**:
I added to PropertyResourceBundle its own static methods getBundle and handleGetBundle to that hide the abstract superclass versions, and implement the proper behavior (load as a .properties file always, and never try to load as a class).

### DIFF
--- a/xalan/third_party/android/platform/external/apache-xml/src/main/java/org/apache/xml/serializer/NamespaceMappings.java
+++ b/xalan/third_party/android/platform/external/apache-xml/src/main/java/org/apache/xml/serializer/NamespaceMappings.java
@@ -25,6 +25,7 @@ import java.util.Hashtable;
 
 import org.xml.sax.ContentHandler;
 import org.xml.sax.SAXException;
+import com.google.j2objc.annotations.Weak;
 
 /**
  * This class keeps track of the currently defined namespaces. Conceptually the
@@ -77,6 +78,7 @@ public class NamespaceMappings
      * mappings at deeper depths push later on such a stack.  Mappings pushed
      * earlier on the stack will have smaller values for MappingRecord.m_declarationDepth.
      */
+    @Weak
     private Hashtable m_namespaces = new Hashtable();
 
     /** 
@@ -92,6 +94,7 @@ public class NamespaceMappings
      * removed at the same time by using this stack to
      * ensure prefix/uri map scopes are closed correctly.
      */
+    @Weak
     private Stack m_nodeStack = new Stack();
 
     private static final String EMPTYSTRING = "";
--- a/jre_emul/android/libcore/luni/src/main/java/java/util/PropertyResourceBundle.java
+++ b/jre_emul/android/libcore/luni/src/main/java/java/util/PropertyResourceBundle.java
@@ -19,7 +19,10 @@ package java.util;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.io.Reader;
+import static java.nio.charset.Charsets.UTF_8;
+import libcore.io.IoUtils;
 
 /**
  * {@code PropertyResourceBundle} loads resources from an {@code InputStream}. All resources are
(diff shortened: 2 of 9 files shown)


gistya:
I signed it!

kstanger:
Thanks for the detailed explanation of the changes. Before getting into too much detail, I am confused by the assumption that an exception being thrown will cause a memory leak. This should not be the case unless the code is compiled with ARC and without using -fobjc-arc-exceptions. Our Xalan library (as with all the libs we distribute) is built without ARC, so thrown exceptions should not result in memory leaks. (otherwise, we'd have many more problems than the ones identified here!)

Here's some [info about why we discourage using ARC with J2ObjC translated code](http://j2objc.org/docs/Memory-Management.html#automatic-reference-counting-arc). And here's more [info about our non-ARC memory model](http://j2objc.org/docs/Memory-Model.html), which doesn't retain local variables.

Can you explain how the exceptions described result in memory leaks?

xalan/third_party/android/platform/external/apache-xml/src/main/java/org/apache/xml/serializer/NamespaceMappings.java
```diff
@@ -77,6 +78,7 @@
      * mappings at deeper depths push later on such a stack.  Mappings pushed
      * earlier on the stack will have smaller values for MappingRecord.m_declarationDepth.
      */
+    @Weak
```
kstanger: This usage of @Weak will be problematic. Marking the collection as weak is not quite the same thing as having a collection with weak references to its elements. If the instance of NamespaceMappings outlives the current autorelease pool, it's m_namespaces and m_nodeStack fields will become dangling pointers. Is there a better place to cut the reference cycle in these cases?

tomball:
@gistya ping -- any response to Keith's question?

tomball:
Closed due to code review not being finished. Please reopen if you have time to finish it.

PR #738, opened 2016-04-15
Apple TV support +  OS X target to 10.7

gianlucabertani:
This PR introduces support for tvOS (Apple TV):

* Two new architectures are added in the J2OBJC_ARCHS: appletvos and appletvsimulator.
* A new output dir. is added, to contain binaries for tvOS builds. This dir. is handled similarly to the OS X case.

Additionally, this PR sets a target for OS X to 10.7, roughly equivalent (in terms of APIs) to the iOS 5.0 target already in place.

Changes have been tested and we are currently using the tvOS binaries with no issues so far.

### DIFF
--- a/make/common.mk
+++ b/make/common.mk
@@ -42,6 +42,11 @@ ARCH_BUILD_MACOSX_DIR = $(ARCH_BUILD_DIR)/macosx
 ARCH_LIB_MACOSX_DIR = $(ARCH_LIB_DIR)/macosx
 DIST_LIB_MACOSX_DIR = $(DIST_LIB_DIR)/macosx
 
+# Appletv library dirs.
+ARCH_BUILD_TV_DIR = $(ARCH_BUILD_DIR)/appletvos
+ARCH_LIB_TV_DIR = $(ARCH_LIB_DIR)/appletvos
+DIST_LIB_TV_DIR = $(DIST_LIB_DIR)/appletvos
+
 ifndef GEN_OBJC_DIR
 GEN_OBJC_DIR = $(BUILD_DIR)/objc
 endif
@@ -50,7 +55,7 @@ GEN_JAVA_DIR = $(BUILD_DIR)/java
 endif
 
 ifndef J2OBJC_ARCHS
-J2OBJC_ARCHS = macosx iphone iphone64 watchv7k simulator simulator64
+J2OBJC_ARCHS = macosx iphone iphone64 watchv7k simulator simulator64 appletvos appletvsimulator
 endif
 
 # xcrun finds a specified tool in the current SDK /usr/bin directory.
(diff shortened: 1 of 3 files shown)


make/fat_lib_macros.mk
```diff
@@ -20,8 +20,10 @@ FAT_LIB_MACOSX_SDK_DIR := $(shell bash $(J2OBJC_ROOT)/scripts/sysroot_path.sh)
 FAT_LIB_IPHONE_SDK_DIR := $(shell bash $(J2OBJC_ROOT)/scripts/sysroot_path.sh --iphoneos)
 FAT_LIB_SIMULATOR_SDK_DIR := $(shell bash $(J2OBJC_ROOT)/scripts/sysroot_path.sh --iphonesimulator)
 FAT_LIB_WATCH_SDK_DIR := $(shell bash $(J2OBJC_ROOT)/scripts/sysroot_path.sh --watchos)
+FAT_LIB_TV_SDK_DIR := $(shell bash $(J2OBJC_ROOT)/scripts/sysroot_path.sh --appletvos)
+FAT_LIB_TVSIMULATOR_SDK_DIR := $(shell bash $(J2OBJC_ROOT)/scripts/sysroot_path.sh --appletvsimulator)
 
-FAT_LIB_MACOSX_FLAGS = $(FAT_LIB_OSX_FLAGS) -DJ2OBJC_BUILD_ARCH=x86_64 \
+FAT_LIB_MACOSX_FLAGS = $(FAT_LIB_OSX_FLAGS) -DJ2OBJC_BUILD_ARCH=x86_64 -mmacosx-version-min=10.7 \
```
kstanger: Our current macosx build seems fine with the current flags. What problem does this solve?

make/common.mk
```diff
@@ -50,7 +55,7 @@ GEN_JAVA_DIR = $(BUILD_DIR)/java
 endif
 
 ifndef J2OBJC_ARCHS
-J2OBJC_ARCHS = macosx iphone iphone64 watchv7k simulator simulator64
+J2OBJC_ARCHS = macosx iphone iphone64 watchv7k simulator simulator64 appletvos appletvsimulator
```
kstanger: Normally our focus has been on iOS builds with the macosx included mainly for testing purposes.
@tomball What are your thoughts on supporting Apple TV apps?

make/fat_lib_macros.mk
gianlucabertani: Building an OS X application that makes use of J2ObjC libs and legitimately targets a previous version of OS X, produces lots of warnings such as:

```
ld: warning: object file (/Users/gbertani/[REDACTED]/j2objc/lib/macosx/libjre_emul.a(BufferedOutputStream.o)) was built for newer OSX version (10.11) than being linked (10.9)
ld: warning: object file (/Users/gbertani/[REDACTED]/j2objc/lib/macosx/libjre_emul.a(BufferedReader.o)) was built for newer OSX version (10.11) than being linked (10.9)
ld: warning: object file (/Users/gbertani/[REDACTED]/j2objc/lib/macosx/libjre_emul.a(BufferedWriter.o)) was built for newer OSX version (10.11) than being linked (10.9)
```

There's no reason to build J2ObjC targeting the latest OS X version if code is compatible with OS X APIs down to 10.7. This flag just lowers the target to an API level comparable to that targeted for iOS.

kstanger: Makes sense, thanks.

make/common.mk
kstanger: We're going to accept the contribution. If we encounter any problems with having appletv builds in the distribution we can always update the J2OBJC_ARCHS variable without touching the rest of the contribution.

kstanger:
Thanks for the contribution!

gianlucabertani:
Thank you for this great project!

PR #742, opened 2016-04-28
Don't catch a generic Exception following a multi-catch statement

danieldickison:
Currently a multi-catch statement like:

    catch (A | B e) {}

is translated to _three_ objc `@catch` clauses:

    @catch (A *e) {}
    @catch (B *e) {}
    @catch (JavaLangRuntimeException *e) {}

The last `@catch` clause seems to be the closest common ancestor of `A` and `B`, but from what I understand, this doesn't match the semantics of the Java multi-catch syntax. In fact, if the multi-catch is followed by a `catch (C e) {}`, where `C` also inherits `RuntimeException`, the translated `@catch (C *e)` clause is never reached because it is shadowed by the extraneous `@catch` clause.

This pull request changes the translation to omit the extra `@catch` clause after a multi-catch. Please review because I can't help but think that the extra `@catch` clause was previously being generated intentionally.

### DIFF
--- a/translator/src/test/java/com/google/devtools/j2objc/gen/StatementGeneratorTest.java
+++ b/translator/src/test/java/com/google/devtools/j2objc/gen/StatementGeneratorTest.java
@@ -1399,6 +1399,7 @@ public class StatementGeneratorTest extends GenerationTest {
         "Test", "Test.m");
     assertTranslation(translation, "@catch (Test_FirstException *e) {\n    @throw e;\n  }");
     assertTranslation(translation, "@catch (Test_SecondException *e) {\n    @throw e;\n  }");
+    assertNotInTranslation(translation, "@catch (JavaLangException *e) {\n    @throw e;\n  }");
   }
 
   public void testDifferentTypesInConditionalExpression() throws IOException {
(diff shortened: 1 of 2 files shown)


translator/src/test/java/com/google/devtools/j2objc/gen/StatementGeneratorTest.java
```diff
@@ -132,6 +132,20 @@ public void testCatchTranslation() throws IOException {
     assertEquals("@try {\n;\n}\n @catch (JavaLangException *e) {\n}", result);
   }
 
+  public void testMultiCatchTranslation() throws IOException {
```
kstanger: Could you instead merge this into testMultiCatch() below, and use assertTranslatedLines to verify the translation.

danieldickison: Good call, I had overlooked `testMutiCatch()` earlier. Updated the pull request but used `assertNotInTranslation()` instead of `assertTranslatedLines()` for the sake of brevity. Do you think that's sufficient?

kstanger:
The Travis build timed out, but at least got past j2objc compilation and translation of jre sources, so I'll pull this anyways.

danieldickison:
Thanks for the quick response and merge!

PR #770, opened 2016-07-20
HttpURLConnection.getHeaderField() should be case-insensitive

danieldickison:
This fixes a bug where `HttpUrlConnection.getHeaderField(String)` was comparing keys in a case-sensitive way. I didn't see any existing tests for this class so I got lazy about creating one for this small fix.

### DIFF
--- a/jre_emul/Classes/com/google/j2objc/net/IosHttpURLConnection.java
+++ b/jre_emul/Classes/com/google/j2objc/net/IosHttpURLConnection.java
@@ -598,7 +598,7 @@ willPerformHTTPRedirection:(NSHTTPURLResponse *)response
 
   private void setHeader(String k, String v) {
     for (HeaderEntry entry : headers) {
-      if (entry.key == k) {
+      if (entry.key == k || (entry.key != null && k != null && k.equalsIgnoreCase(entry.key))) {
         headers.remove(entry);
         break;
       }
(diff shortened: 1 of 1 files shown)


tomball:
Looks good, thanks for the fix!

tomball:
Do you see any other header comparisons that are case-sensitive and shouldn't be?

danieldickison:
@tomball I can think of 2 potential issues:

1. `setRequestProperty(String, String)` calls the private `setHeader(String, String)` which uses `==` to find and remove an existing header with the same key. This will break if the keys are not interned or if they vary in case.
2. `getHeaderFields()` creates a `Map<String, List<String>>` coalescing multiple headers for the same key. That should ostensibly coalesce headers that vary only in case, but this doesn't happen in the current implementation. I think this is debatable: perhaps the API is intended to directly show what the request headers looked like, in which case preserving case in the Map keys makes sense.

Shall I fix at least the first issue in this patch?

tomball:
Yes, please fix the first issue (good catch!). The second issue should be easily fixed by initializing the map with "new TreeMap<>(String.CASE_INSENSITIVE_ORDER)" -- we don't care about the map's order, but keys that vary only by case will be coalesced. And if you don't have time, feel free to push back and I'll make these changes.

danieldickison:
@tomball I fixed the `setRequestProperty()` issue.

I'm going to leave the `getHeaderFields()` to you because `String.CASE_INSENSITIVE_ORDER` will not work here due to there being a `null` key, so it will require (slightly) more code to handle.

tomball:
Thanks for your contribution. After this gets pulled, I'll investigate whether a TreeMap + custom comparator will work (but no promises).

danieldickison:
Thanks. The getHeaderFields() issue seems low-priority since the documentation doesn't make any claims about the case of keys. Probably the only reasonable thing to do is to match the behavior of Android's HttpURLConnection, which I have not investigated.

PR #781, opened 2016-08-16
Add package prefixed filename

phobe77:
Hi,

this adds an option --package-prefixed-filenames which will mangle the output filename to FooBarTest for a java class foo.bar.Test . It only works in combination with --no-package-directories.

Although i'm unable to write proper unit-tests things seem to be alright, j2objc itself compiles and the generated output compiles as well (in my case(s)).

Regards,
## 

Michel

### DIFF
--- a/translator/src/main/java/com/google/devtools/j2objc/Options.java
+++ b/translator/src/main/java/com/google/devtools/j2objc/Options.java
@@ -94,6 +94,7 @@ public class Options {
   private boolean nullability = false;
   private EnumSet<LintOption> lintOptions = EnumSet.noneOf(LintOption.class);
   private boolean includeGeneratedSources = false;
+  private boolean packagePrefixedFilenames = false;
 
   private PackagePrefixes packagePrefixes = new PackagePrefixes();
 
@@ -428,6 +429,8 @@ public class Options {
         outputStyle = OutputStyleOption.SOURCE_COMBINED;
       } else if (arg.equals("-XincludeGeneratedSources")) {
         includeGeneratedSources = true;
+      } else if (arg.equals("--package-prefixed-filenames")) {
+        packagePrefixedFilenames = true;
       } else if (arg.equals("-use-arc")) {
         checkMemoryManagementOption(MemoryManagementOption.ARC);
       } else if (arg.equals("-g")) {
@@ -531,8 +534,8 @@ public class Options {
     }
 
     if (shouldMapHeaders() && buildClosure) {
-      ErrorUtil.error(
-          "--build-closure is not supported with -XcombineJars or --preserve-full-paths");
+      ErrorUtil
+          .error("--build-closure is not supported with -XcombineJars or --preserve-full-paths");
     }
 
     if (memoryManagementOption == null) {
@@ -544,16 +547,22 @@ public class Options {
       nullability = true;
     }
 
+    if (packagePrefixedFilenames && (outputStyle != OutputStyleOption.NONE)) {
+      ErrorUtil.error(
+          "--package-prefixed-filenames is only supported in combination with --no-package-directories");
+    }
+
     // Pull source version from system properties if it is not passed with -source flag.
     if (sourceVersion == null) {
       sourceVersion = SourceVersion.parse(System.getProperty("java.version").substring(0, 3));
     }
 
     // Java 6 had a 1G max heap limit, removed in Java 7.
-    if (batchTranslateMaximum == -1) {  // Not set by flag.
+    if (batchTranslateMaximum == -1) { // Not set by flag.
       batchTranslateMaximum = SourceVersion.java7Minimum(sourceVersion) ? 300 : 0;
     }
 
+    // Package prefixed filenames is only supported in combination with
     int nFiles = args.length - nArg;
     String[] files = new String[nFiles];
     for (int i = 0; i < nFiles; i++) {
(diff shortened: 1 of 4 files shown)


translator/src/main/java/com/google/devtools/j2objc/Options.java
```diff
@@ -94,6 +94,7 @@
   private boolean nullability = false;
   private EnumSet<LintOption> lintOptions = EnumSet.noneOf(LintOption.class);
   private boolean includeGeneratedSources = false;
+  private boolean packagePrefixedFilenames = false;
```
kstanger: Instead of an extra boolean field, use outputStyle and add an enum value to OutputStyleOptions.

translator/src/main/java/com/google/devtools/j2objc/Options.java
```diff
@@ -428,6 +429,8 @@ public static void reset() {
         outputStyle = OutputStyleOption.SOURCE_COMBINED;
       } else if (arg.equals("-XincludeGeneratedSources")) {
         includeGeneratedSources = true;
+      } else if (arg.equals("--package-prefixed-filenames")) {
```
kstanger: Move up next to --preserve-full-paths

translator/src/main/java/com/google/devtools/j2objc/Options.java
```diff
@@ -531,8 +534,8 @@ else if (arg.equals("--final-methods-as-functions")
     }
 
     if (shouldMapHeaders() && buildClosure) {
-      ErrorUtil.error(
-          "--build-closure is not supported with -XcombineJars or --preserve-full-paths");
+      ErrorUtil
+          .error("--build-closure is not supported with -XcombineJars or --preserve-full-paths");
```
kstanger: Revert this diff. Having the line break after the "(" is more in line with internal Google style.

translator/src/main/java/com/google/devtools/j2objc/Options.java
```diff
@@ -544,13 +547,19 @@ else if (arg.equals("--final-methods-as-functions")
...
+    }
+
     // Pull source version from system properties if it is not passed with -source flag.
     if (sourceVersion == null) {
       sourceVersion = SourceVersion.parse(System.getProperty("java.version").substring(0, 3));
     }
 
     // Java 6 had a 1G max heap limit, removed in Java 7.
-    if (batchTranslateMaximum == -1) {  // Not set by flag.
+    if (batchTranslateMaximum == -1) { // Not set by flag.
```
kstanger: Revert this diff. Our internal style prefers the two spaces before the comment.

translator/src/main/java/com/google/devtools/j2objc/Options.java
```diff
@@ -713,17 +722,19 @@ public static File getOutputDirectory() {
...
   }
 
+  public static boolean generatePackagePrefixedFilenames() {
+    return instance.packagePrefixedFilenames;
+  }
+
   /**
-   * If true, put output files in sub-directories defined by
-   * package declaration (like javac does).
+   * If true, put output files in sub-directories defined by package declaration (like javac does).
```
kstanger: Please revert all other extraneous diffs, since they just clutter the commit.

translator/src/main/java/com/google/devtools/j2objc/gen/GenerationUnit.java
```diff
@@ -206,16 +205,19 @@ public boolean isFullyParsed() {
...
    */
   private static String getDefaultOutputPath(CompilationUnit unit) {
     String path = unit.getMainTypeName();
     PackageDeclaration pkg = unit.getPackage();
     if (Options.usePackageDirectories() && !pkg.isDefaultPackage()) {
       path = pkg.getName().getFullyQualifiedName().replace('.', File.separatorChar)
           + File.separatorChar + path;
+    } else if (Options.generatePackagePrefixedFilenames()) {
+      // Prepend the package in camelcase to the class
+      path = NameTable.camelCaseQualifiedName(pkg.getName().getFullyQualifiedName()) + path;
```
kstanger: Use nameTable.getPrefix(pkg). This will incorporate any package prefix renaming into your file name.

kpgalligan:
Haven't seen any updates on this for a while, so I took a whack at it. Applied changes and feedback. Seems to work well with my code, although tests need to be augmented somewhat. I applied changes to a fork, so would need to apply to a clean clone before doing a pull request, but see changes here: https://github.com/doppllib/j2objc/commit/415a599c4310043e45cc2361e5d154e08b2bd93d

phobe77:
Keith,

i completely missed this one somehow. At this point, do you need me to do anything?

kstanger:
Let me get back to you both on this. I'd like to do some refactoring to clean up some of the related code. Ideally, the code for generating output and include file names would be contained within the OutputStyleOptions enum. That way we can easily add additional ways to generate outputs without littering the code with if-else blocks.
Keep pinging me if you don't see action on this in a few weeks.

kstanger:
I did some refactoring in [this commit](https://github.com/google/j2objc/commit/78b971530a4ec6089b0dfeee728de76c710d26a6). Now most of the changes required for this new option should be limited to [this switch statement in HeaderMap](https://github.com/google/j2objc/blob/master/translator/src/main/java/com/google/devtools/j2objc/util/HeaderMap.java#L195), plus adding the new OutputStyleOptions enum value.

AlexeyBelezeko:
Hi,
As I understand this pull request is no longer needed by the author. Should I make my own pull request to contribute this feature?
I've also found that adding of prefix to file name and import in implementation part can't be limited just to a switching statement in HeaderMap.

phobe77:
Hi Alexey,

I'm not able to work on this pull at the moment. So if you feel like taking it over I'm all for it.

-- 
Michel

tomball:
Alexey, please either reopen or send a separate pull request if you have time to finish this work.

Issue #785, opened 2016-08-31
Add Prefixes to Category Methods on NSObject


efirestone:
We recently ran into an issue where `+[NSObject getClass]` provided by the `NSObject (JavaObject)` category was getting called unintentionally from our non-J2ObjC code. The single shared namespacing in Objective-C means that category methods are available to all code in the memory space, and with a category on NSObject, it means it applies to all Objective-C objects.

[Apple recommends prefixing methods](https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/CustomizingExistingClasses/CustomizingExistingClasses.html#//apple_ref/doc/uid/TP40011210-CH6-SW4) that you add in a category [[more here](http://nshipster.com/namespacing/)], and that would solve this issue as well. For example, changing `+[NSObject getClass]` to `+[NSObject java_getClass]` or `+[NSObject J2OBJC_getClass]`.

This is most important for the `NSObject` category, but applies equally well to any category provided by J2ObjC.

tomball:
Sounds reasonable. My vote is for "java_" prefixes, since it's more obvious why the method is necessary. I can imagine more "why are there j2objc methods in this category?" whining than "why are there java methods in this category?" :-)

efirestone:
Thanks @tomball, working on a fix now. It looks like `finalize` has actually already been renamed to `javaFinalize` so I'll go with that pattern (`javaClone`, `javaGetClass`, etc) unless you would prefer we switch to something else (at which point I can change `javaFinalize` as well).

tomball:
That sounds great, thanks!

tomball:
All NSObject+JavaObject methods now have java_ prefixes, as do the ones in the other jre_emul categories.

PR #787, opened 2016-09-01
Prefix categories on NSObject to avoid potential conflicts

efirestone:
https://github.com/google/j2objc/issues/785


efirestone:
I signed it!

translator/src/main/resources/com/google/devtools/j2objc/JRE.mappings
```diff
@@ -98,12 +104,18 @@ java.lang.Class.isInterface()Z = isInterface
...
+# java.lang.Comparable
+java.lang.Comparable.compareTo(Ljava/lang/Object;)I = javaCompareTo:
+
 # java.lang.String
 java.lang.String.charAt(I)C = charAtWithInt:
 java.lang.String.codePointAt(I)I = codePointAt:
 java.lang.String.codePointBefore(I)I = codePointBefore:
 java.lang.String.codePointCount(II)I = codePointCount:endIndex:
-java.lang.String.compareTo(Ljava/lang/String;)I = compareToWithId:
+java.lang.String.compareTo(Ljava/lang/String;)I = javaCompareTo:
```
kstanger: I would expect you can remove this line without consequence because of the Comparable.compareTo mapping above.

efirestone: I had just updated the existing entry without thinking about it, but looks like you're right. Removed.

kstanger:
@tomball this looks good to me. But we should be careful because the API change could break internal users when we merge it in. We may want to temporarily leave the original NSObject category methods and have them delegate to the new methods until we can migrate user call sites.

tomball:
I'll do a code search tomorrow to see if there's an issue, and if so,
proceed as you suggest.

tomball:
Sorry for the delay merging this pull request, but we haven't forgotten it. Monday we're staging a big chunk of the Java 8 API (streams, functions, etc.), and making it public once we verify nothing broke. We'll should be able to pull this request next week.

danieldickison:
I think this is a good idea and I think it should go further to adopt the prefix on categories of other Foundation types. E.g. in`NSString+JavaString.h/m`.

(As an aside, while I don't think this needs to be a motivating factor, a conflict on category method `[NSString split:]` was at least one of the issues that prevented a lib built with j2objc from working in a RubyMotion app when I evaluated that last year.)

Also, I've seen other projects use a prefix with an underscore to reduce the chances of conflict even more, I.e. something like `java_compareTo:`

tomball:
I agree that moving to using a prefix everywhere is the best strategy.
Since Xcode code-completion makes it very easy to write native code that
ties to the current API, I think we should update one category at a time
and leave some time in between for app developers to recover.

kstanger:
For the record, I also like the underscore naming scheme.

efirestone:
The underscore is the recommendation from Apple I think. I was going to go that way originally, but went with the style that already existed. Happy to switch it, just let me know.

tomball:
Yes, please add the underscore so we're fully following Apple's recommendations. Once you do, we are now at a point where this can be pulled.

efirestone:
Prefixes have been updated.

jre_emul/Classes/NSString+JavaString.m
```diff
@@ -1005,6 +1004,7 @@ + (const J2ObjcClassInfo *)__metadata {
     { "indexOfString:fromIndex:", "I", 0x1, 45, 46, -1, -1, -1, -1 },
     { "intern", "LNSString;", 0x1, -1, -1, -1, -1, -1, -1 },
     { "isEmpty", "Z", 0x1, -1, -1, -1, -1, -1, -1 },
+    { "java_compareTo:", "I", 0x1, 33, 13, -1, -1, -1, -1 },
```
efirestone: Can someone more familiar with the way J2ObjC reflection works verify that changing the ordering here isn't going to throw any indexes off?

tomball: Order doesn't matter. Most classes' method lists are short enough that a
linear comparison is fast enough, and large apps have so many classes that
hash tables would take too much memory. Thanks for checking, though.

kstanger: Here's the script we use to update the metadata in our native-written classes. As Tom mentions the ordering shouldn't matter, but by using the script you can be sure the metadata is correct and prevent a future diff the next time Tom or I run it.
https://github.com/google/j2objc/blob/master/jre_emul/update_metadata.sh

ifabijanovic:
Hello, any progress on this one? It is a blocker for our project

tomball:
Eric seems to be busy, so I've started this migration. I'm doing
Object.getClass() first, since it impacts several project files and one of
our apps. I'll follow that with the rest of NSObject+JavaObject.
NSString+JavaString next, since it's been reported
<https://github.com/google/j2objc/issues/798#issuecomment-253209252> that
one of iOS 10's frameworks also added an intern method to NSString.

On Wed, Oct 19, 2016 at 6:54 AM Ivan Fabijanović <notifications@github.com>
wrote:

Hello, any progress on this one? It is a blocker for our project

—
You are receiving this because you were mentioned.


Reply to this email directly, view it on GitHub
<https://github.com/google/j2objc/pull/787#issuecomment-254819397>, or mute
the thread

ifabijanovic:
Hello, I see some work on this has been merged to the `master` branch. I checked it out, built using `make frameworks` and ran tests using `make test`. I got 8 failures:

Tests run: 6498,  Failures: 8

make[1]: *** [run-tests] Error 1
make: *** [test_jre_emul] Error 2

I am unsure if these are all the tests or was testing interrupted because of failed test. Any advice on how to continue from here? We still haven't integrated j2objc to our production because of this issue.

kstanger:
The tests shouldn't be failing. If they are, it's likely an unrelated issue. Please file a separate issue and include the details of the failed tests.

tomball:
A "java_" prefix has been added to the methods in NSObject+JavaObject and NSString+JavaString. We've deferred prefixing the methods in NSException+JavaThrowable, because we're still debating whether it's better if java.lang.Throwable extends NSException instead of being mapped to it, and if that's done, how to support a "catch everything" try statement in Java.

Eric, thanks for your push to make this happen in a timely manner. Although we didn't use the actual push code, you are the driver behind this enhancement, and we appreciate your help.

PR #789, opened 2016-09-09
update Xcode Build Rules documentation

theRealWardo:
- update setup to explain use of an `J2OBJC_HOME` variable to make tweaking incorrect paths easier. (this is closer to the sample project mentioned at the top of the document.)

- updates screenshots to Xcode 7

- clarifies that linking additional libraries is not always necessary

- adds more debugging tips

preview this change at https://github.com/theRealWardo/j2objc/blob/6d962f8d18336189455506a91878a948b02787a6/docs/Xcode-Build-Rules.md


tomball:
Thank you very much for the rewrite. You have a future as a doc-writer if you ever get tired of coding!

PR #824, opened 2016-12-22
Don't redeclare generic private method in subclass

danieldickison:
If a superclass contains a method that returns a parameterized type, a subclass that narrows that type shouldn't include a declaration of that method in the .m file if it is a private method since the subclass is restricted from calling that method. In practice, this was causing compilation errors if the method argument list contained types that are not already imported in the subclass.

This patch fixes the issue and includes a test case.

### DIFF
--- a/translator/src/main/java/com/google/devtools/j2objc/translate/AbstractMethodRewriter.java
+++ b/translator/src/main/java/com/google/devtools/j2objc/translate/AbstractMethodRewriter.java
@@ -129,7 +129,7 @@ public class AbstractMethodRewriter extends UnitTreeVisitor {
     for (DeclaredType inheritedType : typeUtil.getObjcOrderedInheritedTypes(type.asType())) {
       TypeElement inheritedElem = (TypeElement) inheritedType.asElement();
       for (ExecutableElement methodElem : ElementUtil.getMethods(inheritedElem)) {
-        if (methodElem.getModifiers().contains(Modifier.PRIVATE)) {
+        if (ElementUtil.isPrivate(methodElem)) {
           continue;
         }
         TypeMirror declaredReturnType = typeUtil.erasure(methodElem.getReturnType());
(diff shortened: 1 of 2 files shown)


translator/src/main/java/com/google/devtools/j2objc/translate/AbstractMethodRewriter.java
```diff
@@ -129,6 +129,9 @@ private void addReturnTypeNarrowingDeclarations(AbstractTypeDeclaration node) {
     for (DeclaredType inheritedType : typeUtil.getObjcOrderedInheritedTypes(type.asType())) {
       TypeElement inheritedElem = (TypeElement) inheritedType.asElement();
       for (ExecutableElement methodElem : ElementUtil.getMethods(inheritedElem)) {
+        if (methodElem.getModifiers().contains(Modifier.PRIVATE)) {
```
kstanger: Please use "ElementUtil.isPrivate(methodElem)".

kstanger:
Looks good with one comment to fix.

kstanger:
@danieldickison do you think you could provide a test case that reproduces the compile error mentioned? That seems like it might be a bug that's not specific to private methods but I can't reproduce it. In my test I have a "@class" declaration for the parameter type as expected.

danieldickison:
@kstanger I failed to come up with a test case that reproduces that error. I see it happening in a particular class in our app's actual code base using j2objc 1.2, but there must be other conditions that cause the bug to appear.

What I see is that with our app's code, using the 1.2 distribution of j2objc, the subclass gains the method declaration in the .m file's "private" `@interface` block, but it doesn't add a `@class` declaration for the argument type `JavaIoIOException`, causing an objc compilation error. However, when I run the `testGenericPrivateMethodNotAdded` test case against the 1.2 tag, I find that the subclass .m file does _not_ contain the method declaration, or an `@interface` block at all. So it seems I haven't reproduced the spurious private method declaration bug against 1.2 after all.

On the other hand, against the current master branch, `testGenericPrivateMethodNotAdded` _does_ add the spurious method declaration in the subclass .m file. However, when I add an argument to that method, it also adds the `@class` for the argument type, thereby avoiding the error (as you saw).

So I'm not sure what's causing the bug with our app code but I don't think I'll have the time to investigate this much further. I do think the change in this pull request is still valid, though let me know if you disagree. (I've made the ElementUtil.isPrivate() change as requested.)

tomball:
Thanks for taking the extra time with this.

kstanger:
It's a good change and a good test addition. Thanks for the contribution! It doesn't seem worth extra effort to track down the potential type declaration issue.

PR #839, opened 2017-03-08
Fix for IosHttpURLConnection not validating SSL certificates

smdvdsn:
IosHttpURLConnection currently does not validate SSL certificates.
This change updates the URLSession didReceiveChallenge delegate method
to complete with NSURLSessionAuthChallengePerformDefaultHandling to
allow iOS to handle TLS challenge.


tomball:
Please review, Lukhnos, as you're more familiar with NSURLSession than me. When okay, please reassign this issue to me to pull it.

lukhnos:
LGTM. Thanks for the patch!

PR #840, opened 2017-03-08
IosHttpURLConnection - stop NSURLSession caching to disk when useCaches is false     

smdvdsn:
NSURLSession has a response cache that is on by default. This change disables the
cache when HttpURLConnection.useCaches is false.

This pull came about from a need to stop http responses being written to device storage
for security reasons.


tomball:
Looks good to me. Assigning to Lukhnos for review, and if okay, please reassign to me to pull it.

lukhnos:
LGTM. Back to @tomball.

PR #852, opened 2017-04-27
Fixed several issues of SystemClock.

th3hamm0r:
As an Android developer I was very happy to use `SystemClock` in the shared code for a media playing app which heavily relies on timers. Unfortunately I ran into the following issues:

* The iOS implementation of `elapsedRealtime()` is not monotonic, as the used `systemUptime` does not take sleep time into account.
The issue was hard to find, since it works in many cases and I was blind, knowing the contract of the Android's API ;)
* `elapsedRealtimeNanos()` returns the same value during the lifetime of a process, since it only returns the actual boot time of the system. But as the SO post below describes, even this fixed time can change, if the user/system changes the OS time.
The implementation of `elapsedRealtimeNanos()` actually has another issue: the returned value contains a timestamp in microseconds.

In my change I took the code of `elapsedRealtimeNanos()` to get the boot time in microseconds and used it to calculate the actual uptime based on this SO post: http://stackoverflow.com/q/12488481/1394330
I also took the warning of a race condition into account. `elapsedRealtimeNanos()` still uses the microseconds multiplied by 1000 to calculate the nanoseconds. This may be improved by using other APIs (?), but for now I'm ok with this ;)

### DIFF
--- a/jre_emul/android/frameworks/base/core/java/android/os/SystemClock.java
+++ b/jre_emul/android/frameworks/base/core/java/android/os/SystemClock.java
@@ -138,21 +138,8 @@ public final class SystemClock {
       return elapsedRealtime();  // Not aware of any iOS API to track deep sleep time.
     }
 
-    /**
-     * Returns milliseconds since boot, including time spent in sleep.
-     *
-     * @return elapsed milliseconds since boot.
-     */
-    public native static long elapsedRealtime() /*-[
-      return (long long) [[NSProcessInfo processInfo] systemUptime] * 1000LL;
-    ]-*/;
-
-    /**
-     * Returns nanoseconds since boot, including time spent in sleep.
-     *
-     * @return elapsed nanoseconds since boot.
-     */
-    public static native long elapsedRealtimeNanos() /*-[
+    /*-[
+    static long long boottime_us() {
       // MIB = Management Information Base (man sysctl).
       #define MIB_SIZE 2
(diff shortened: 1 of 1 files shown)


th3hamm0r:
I signed it!

jre_emul/android/frameworks/base/core/java/android/os/SystemClock.java
```diff
@@ -138,21 +138,8 @@ public static long uptimeMillis() {
...
-    ]-*/;
-
-    /**
-     * Returns nanoseconds since boot, including time spent in sleep.
-     *
-     * @return elapsed nanoseconds since boot.
-     */
-    public static native long elapsedRealtimeNanos() /*-[
+    /*-[
+    static long long boottime_us() {
```
tomball: Please change to boottime_usec to aid readability.

jre_emul/android/frameworks/base/core/java/android/os/SystemClock.java
```diff
@@ -167,6 +154,40 @@ public static native long elapsedRealtimeNanos() /*-[
                             initWithNSString:@"sysctl" withInt:errno]);
       }
       return ((long long) boottime.tv_sec) * 1.e6 + boottime.tv_usec;
+    }
+
+    static long long uptime_us() {
```
tomball: Please change to uptime_usec().

PR #860, opened 2017-05-29
Fix retain cycle by making Reader have @Weak reference to lock

danieldickison:
I believe there's a retain cycle leak in `Reader.lock`. I'm pretty sure this is the case when `this` is used as the lock, since a strong reference to itself is always a retain cycle unless broken elsewhere. In the case of `StreamDecoder` (a subclass of `Reader`), when used by `InputStreamReader` the ISR instance is used as the StreamDecoder's lock, while the ISR retains a strong reference to the StreamDecoder resulting in a retain cycle.

However, I'm not 100% sure if `@Weak` is safe to use here as I haven't yet audited every use of the lock constructor parameter and field. I could use some outside expertise in this case, since I'm not sure I have the time to check every caller.

### DIFF
--- a/jre_emul/android/platform/libcore/ojluni/src/main/java/java/io/Reader.java
+++ b/jre_emul/android/platform/libcore/ojluni/src/main/java/java/io/Reader.java
@@ -25,6 +25,7 @@
 
 package java.io;
 
+import com.google.j2objc.annotations.Weak;
 
 /**
  * Abstract class for reading character streams.  The only methods that a
@@ -57,7 +58,7 @@ public abstract class Reader implements Readable, Closeable {
      * the object in this field rather than <tt>this</tt> or a synchronized
      * method.
      */
-    protected Object lock;
+    @Weak protected Object lock;
 
     /**
      * Creates a new character-stream reader whose critical sections will


kstanger:
Coincidentally an internal user reported the same leak just this morning. I already have a fix (which avoids the cycle instead of using `@Weak`. I'll do a repo sync so you can pick it up.
FYI: The fix creates a separate lock object in InputStreamReader to pass in to the StreamDecoder instead of "this".

kstanger:
Here's the fix: https://github.com/google/j2objc/commit/390997f1dda1c07155b17b403fdaa1cda5601c90

danieldickison:
Ah perfect timing! I think there's still an argument to be made that the no-arg constructor shouldn't create a `Reader` instance that leaks, but on the other hand, with the field being protected I think it's fair to make this the responsibility of the subclasser.

kstanger:
Oh, I didn't notice the no-arg constructor in java.io.Reader. I should fix that. Assigning "this.lock = new Object();" should do the trick.

danieldickison:
@kstanger It looks like there are some other locations with a similar `ivar = this` pattern, such as `java.io.Writer`. It seems these are regressions from a previously fixed bug: https://github.com/google/j2objc/pull/606

kstanger:
java.io.Reader and java.io.Writer are fixed here: https://github.com/google/j2objc/commit/76963413ac36a7a11801ddc9edf1b3b750db5a67

tomball:
These are due to the move from Apache Harmony sources to OpenJDK.
Technically these may be considered regressions, but it's not from careless
maintenance.

danieldickison:
I figured the OpenJDK switch was the source. I definitely didn't mean to imply it was from careless maintenance, just wanted to make sure those were on the radar :) Thanks for the fixes!

PR #869, opened 2017-07-20
Fix a self-retain cycle in Throwable in 2.0.2

danieldickison:
The new implementation of `Throwable` in the 2.0.2 release has a retain cycle when initialized without a `cause` because the `cause` instance variable is set to `this`. This changes it to reference a sentinel value instead.


kstanger:
Thanks! Looks good. I'll just wait on the Travis CI build before merging.

PR #900, opened 2017-10-22
cast to void to prevent warning

jaspervp:


translator/src/main/java/com/google/devtools/j2objc/translate/EnumRewriter.java
```diff
@@ -200,9 +202,9 @@ private void addNonArcInitialization(EnumDeclaration node) {
       }
 
       initStatements.add(new ExpressionStatement(new CommaExpression(
-          new Assignment(new SimpleName(varElement), new Assignment(
+          new CastExpression(voidType,new ParenthesizedExpression(new Assignment(new SimpleName(varElement), new Assignment(
```
tomball: Limit line length to 100 characters, use 4 character indent on second line.

tomball:
Thanks for your contribution!

PR #941, opened 2018-02-09
Add all output to a single Objective-C file set. Similar to combineJa…

kpgalligan:
…rs, but can be for everything.

The general goal is to be able to send a whole bunch of java files into one Objc output file set. This is to solve a few problems.

1. Xcode doesn't consider paths distinct. Same file name in different paths is a problem (if added visually or with cocoapods)
2. Our usual combination of dependencies winds up with thousands of artifacts. Xcode seems to handle a single huge file much better than many small.
3. combineJar would work if we jar'd up the java and translated the jar, but you can't run java debug on it (the jar gets unzipped in a temp folder, so the #line directives point to non-existent files).

On the negative side, any change to the a single java file triggers a full rebuild of everything, but the usage flow we've had is to do dev in Java/Android, then translate and work in Swift. Usually not a big deal.

Our gradle plugin and framework had a moderately forked j2objc, but I've reduced that to this PR. If we can do this (or something equivalent), we'll be using a stock J2objc build and just publish a gradle plugin and libraries.

(I have another PR coming, but its just to output a warning suppression line)

### DIFF
--- a/translator/src/test/java/com/google/devtools/j2objc/J2ObjCTest.java
+++ b/translator/src/test/java/com/google/devtools/j2objc/J2ObjCTest.java
@@ -131,6 +131,13 @@ public class J2ObjCTest extends GenerationTest {
     makeAssertionsForJavaFiles(exampleH, exampleM, packageInfoH, packageInfoM);
   }
 
+  private void makeAssertionsForGlobalCombinedOutputJavaFiles() throws Exception {
+    String exampleH = getTranslatedFile("testGlobalCombinedOutputFile.h");
+    String exampleM = getTranslatedFile("testGlobalCombinedOutputFile.m");
+    assertTranslation(exampleM, "#include \"testGlobalCombinedOutputFile.h\"");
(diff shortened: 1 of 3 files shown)


translator/src/test/java/com/google/devtools/j2objc/J2ObjCTest.java
```diff
@@ -151,6 +158,19 @@ private void makeAssertionsForJavaFiles(
     makeAssertions(exampleH, exampleM, packageInfoM);
   }
 
+  private void makeAssertionsForGlobalCombinedOutputJavaFiles(
```
tomball: Please use 4 spaces for method and line continuations (we follow the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html)). Line maximum is 100 columns, so the method parameters can be on the first line, with the throws Exception 4 spaces in from the method's 2 space indent.

kpgalligan: Updated!

tomball:
Thanks for your contribution!

PR #947, opened 2018-02-28
ConcurrentSkipListMap: Use sentinel instead of self references

### LINKED ISSUE #946: Leak in ConcurrentSkipListMap's Marker nodes
Marker nodes are bound to leak in `ConcurrentSkipListMap` as they are self-referencing.

https://github.com/google/j2objc/blob/master/jre_emul/android/platform/libcore/luni/src/main/java/java/util/concurrent/ConcurrentSkipListMap.java#L415

As self referencing go, I believe the proper workaround / fix would either be to weakify the self reference iff this is a marker node, OR to break the cycle when removing a marker node from the ListMap.

flovilmart:
- The self references in the marker nodes are leaking
- as recommended, implement a sentinel instead

closes #946 

cc: @AlexandreHetu

### DIFF
--- a/jre_emul/android/platform/libcore/luni/src/main/java/java/util/concurrent/ConcurrentSkipListMap.java
+++ b/jre_emul/android/platform/libcore/luni/src/main/java/java/util/concurrent/ConcurrentSkipListMap.java
@@ -412,7 +412,7 @@ public class ConcurrentSkipListMap<K,V> extends AbstractMap<K,V>
          */
         Node(Node<K,V> next) {
             this.key = null;
-            this.value = this;
+            this.value = sentinel();
             this.next = next;
         }
 
@@ -474,7 +474,7 @@ public class ConcurrentSkipListMap<K,V> extends AbstractMap<K,V>
              * interference among helping threads.
              */
             if (f == next && this == b.next) {
-                if (f == null || f.value != f) // not already marked
+                if (f == null || f.value != sentinel()) // not already marked
                     casNext(f, new Node<K,V>(f));
                 else
                     b.casNext(this, f.next);
@@ -1085,7 +1085,7 @@ public class ConcurrentSkipListMap<K,V> extends AbstractMap<K,V>
                     n.helpDelete(b, f);
                     break;
                 }
-                if (b.value == null || v == n)      // b is deleted
+                if (b.value == null || v == sentinel())      // b is deleted
                     break;
                 if (f != null) {
                     b = n;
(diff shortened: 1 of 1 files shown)


flovilmart:
I signed it!

flovilmart:
@tomball I'm not sure if I didn't miss any comparison, and it's tough for us to test against our project (we're stuck on the 1.0.2 version).

tomball:
Thanks for your contribution!

flovilmart:
@tomball just a FYI and I feel bad about it, the tests didn't manage to run on my machine 

``` 
clang path: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang
SDK path: /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.13.sdk
j2objcc -c /Users/florent/src/j2objc/jre_emul/build_result/tests/android/test/MoreAsserts.m
/Users/florent/src/j2objc/jre_emul/build_result/tests/android/test/MoreAsserts.m:26:10: fatal error:
      'junit/framework/Assert.h' file not found
#include "junit/framework/Assert.h"
         ^~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
make: *** [/Users/florent/src/j2objc/jre_emul/build_result/tests/android/test/MoreAsserts.o] Error 1
```

I initially opened the PR in hope the CI would pick it up and run it faster that I was able to undestand the issue. If there's anything, let me know.

tomball:
It's not an issue, just run "make -j8 dist" from your j2objc directory, so that all sub-projects are built, especially junit.

flovilmart:
Awesome! I wish in was in the contributing guide (I’ll open a PR later).

I ran `make test` from the jre_emul folder.

flovilmart:
@tomball turns out the tests are failing now, sorry for the bad commit.

tomball:
No worries. I need to roll the change back and attack the problem from
another angle.

The problem is this class has several fields which may outlive the instance
(keySet, entrySet, values, and descendingMap), and so need RetainedWith
annotations to ensure these reference cycles don't keep the whole map from
leaking. I don't understand why the map is now being released early (that's
likely why the test failed), but it should be apparent in the debugger.

flovilmart:
I backported to the 1.0.2 version and tried on our project and reach a similar conclusion, after deallocation, a full chain of nodes was leaked. I’ll keep exploring other avenues in our ‘fork’ and keep you posted in case I find something interesting. I’ll reopen the issue also. Thanks for the quick responses, it’s really appreciated.

PR #954, opened 2018-03-08
Add watchOS simulator support

maxibystro:
Now (before changes from this PR) iOS jre_emul static library is still built with armv7k architecture. This slice should be a part of a separate watchOS jre_emul library that should also contain slice for watchsimulator sdk. Just like it has been done for tvOS.


tomball:
Thanks for your contribution!

PR #955, opened 2018-03-10
ConcurrentSkipListMap: Use sentinel instead of self references

### LINKED ISSUE #946: Leak in ConcurrentSkipListMap's Marker nodes

flovilmart:
- The self references in the marker nodes are leaking
- as recommended, implement a sentinel instead

Adds sentinel to other checks

@tomball this one looks good! I found the source of the NPE's, which ultimately was just a logic issue

https://github.com/google/j2objc/compare/master...flovilmart:fix/946-skipListMap-retain-cycles?expand=1#diff-9ccaf044b31bb6a1c42d738f5c33c238R492

Also found out that more checks were done one `node.value == node`

Closes #946

### DIFF
--- a/jre_emul/android/platform/libcore/luni/src/main/java/java/util/concurrent/ConcurrentSkipListMap.java
+++ b/jre_emul/android/platform/libcore/luni/src/main/java/java/util/concurrent/ConcurrentSkipListMap.java
@@ -405,7 +405,7 @@ public class ConcurrentSkipListMap<K,V> extends AbstractMap<K,V>
 
         /**
          * Creates a new marker node. A marker is distinguished by
-         * having its value field point to itself.  Marker nodes also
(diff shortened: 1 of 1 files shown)


tomball:
Thanks for following up on this!

flovilmart:
No problem @TomBall! We found others and are looking to get a zero leak app :)

PR #959, opened 2018-03-13
Use object to synchronizes instead of this

flovilmart:
@tomball this is not tested yet (in the wild), just opening the PR. This still is sound


tomball:
This should be safe, since it follows [the same pattern that SynchronizedCollection uses](https://github.com/AmpMe/j2objc/blob/4a983dbf084a0dee90bdc565e2b2c859e31e5f8b/jre_emul/android/platform/libcore/ojluni/src/main/java/java/util/Collections.java#L1851).

flovilmart:
We’ve set ourselves correctly for ‘alpha’ releases, so i’ll be reporting if we see any weird behaviors.

flovilmart:
@tomball ran some tests this morning and yesteday, everything look good on our side and we're considering using the master branch for our next release.

Issue #960, opened 2018-03-16
Java - @Nonnull constructors


sarsonj:
Hi,

we are using j2objc together with Swift. By default, J2Objc constructors are not annotated as `@Nonnull`, so that when creating instance of them, the "!" has to be used:

```swift
var worker = MyWorker()!
```

It makes the Swift code silly. It is possible to solve this, by defining constructor in Java:
```java
@Nonnull
@ObjectiveCName("init")
public MyWorker() { }
```

but then there has to be this boilerplate code in every class.

I tried to set package-info.java and set `@ParametersAreNonnullByDefault` that should do this, but it looks like this doesn't work with constructors. Any idea, how to make this more swift friendly?

tomball:
Good question. I think it should be safe to always annotate translated Java constructors as nonnull. Objective C initializers may return null if there's a problem, but Java constructors always return a valid object or throw an exception (constructors can have a return statement, but it can't return a different value than "this").

Can you test a translator source change? In [ElementUtil.isNonnull()](https://github.com/google/j2objc/blob/master/translator/src/main/java/com/google/devtools/j2objc/util/ElementUtil.java#L708), make this change:
```
  public static boolean isNonnull(Element element, boolean parametersNonnullByDefault) {
    return hasNonnullAnnotation(element)
        || isConstructor(element)  // Add this line
        || (isParameter(element)
            && parametersNonnullByDefault
            && !((VariableElement) element).asType().getKind().isPrimitive());
  }
```
If this improves your project's Swift importing without breaking anything (it shouldn't), then I'll commit this change with a unit test.

sarsonj:
Hi,

in the meantime, I created pull request: #961. I choose different approach in source code and respect j2objc settings. Please check it! Thanks.

sarsonj:
Anyway, your one line change works as well and is nicer. At least I had chance to look at j2objc source code :-).

Thanks for quick help!

tomball:
You'll find a lot of these one-definition methods in the com.google.devtools.j2objc.util.*Util classes. Because ASTs and code generators make it easy to duplicate code, we've tried to moved common tests like this into utility methods. This makes it easier to change definitions and ensure they are consistent across all code. In this case, we're changing the definition of what elements should be non-null.

I'm going to close your pull request in favor of the one-liner (plus test), but appreciate you taking the time to investigate this issue and verify its fix.

tomball:
Fixed in current source.

PR #961, opened 2018-03-16
When --nullability option enabled, generate constructors with __nonnu…

sarsonj:
Why?

Currently, all j2objc classes are generated with nullable constructors. So that in Swift, we have to use:

`var myInstance = MyClass()!`

 to let compiler know, that method instance is for sure not null. Because Java class instance AFAIK cannot never return null, it is safe to declare all constructors with __nonnull annotation.

This commit respects options of j2objc (only insert this annotation, when j2objc runs with --nullability and also checks, if is constructor not annotated by Java - in this case, the generated code is untouched.


sarsonj:
I signed it!

tomball:
As discussed in #960, this change won't be pulled in favor of a simpler solution. We still appreciate your contribution!

Issue #984, opened 2018-06-14
Use Path Prefixed import in java/lang/reflect/Field.h


congt:
When importing `AccessibleObject.h` in Field.h, it doesn't include its path (https://github.com/google/j2objc/blob/master/jre_emul/Classes/java/lang/reflect/Field.h). It should be #import "java/lang/reflect/AccessibleObject.h".

PR #985, opened 2018-06-15
#984 Uses the path prefixed import for AccessibleObject.h in Field.h

congt:
AccessibleObject.h is in the package java/lang/reflect. A minor issue found when I convert the JRE.framework to be compatible with cocoapods.


congt:
@googlebot I work for Square, Inc. which has a corp. CLA signed.

tomball:
Thanks for offering this contribution, but I already fixed the problem yesterday and we should have it pushed publicly later today. In the future, please mention in the bug report whether you plan to make a pull request, so we wait for it rather than duplicate your work. We want to encourage contributions, after all!

PR #990, opened 2018-07-27
Fixed Spelling.

jimmycasey:


tomball:
Thank you for your contribution to help make j2objc better, we appreciate your effort.

We cannot pull this request, however, because only one source file is actually authored by us, TranslationProcessor.java (apparently "ocurred" is a universal spelling mistake!). We avoid making changes to those sources that don't explicitly fix bugs, since doing otherwise makes it harder to merge our changes with future versions of those external sources.

I encourage you to break this push request up into separate patches for each organization that created a subset of these files. Here are instructions for submitting patches to them:

1. Android: https://source.android.com/setup/contribute/submit-patches
1. OpenJDK (the `android/*/ojluni/*` classes): http://openjdk.java.net/contribute/
1. ICU: http://site.icu-project.org/processes/contribute
1. Apache Xalan: http://www.apache.org/dev/contributors

jimmycasey:
Thanks for the info.

PR #994, opened 2018-07-31
IOSClass initialize issue

sfuku7:
Hi team,

I found one issue, failure of Class.forName() for a class has package prefix.

I investigated this and finally I could find the root cause.
That is, IOSClass_initialize() is not called at the top of the static methods of java.lang.Class (IOSClass).
I guess IOSClass_initialize() should be called, because other translated code includes such code.

Could you please review this ?

I created two commits. The one is to reproduce the issue, and the other one is the fix.

### DIFF
--- a/jre_emul/Classes/IOSClass.m
+++ b/jre_emul/Classes/IOSClass.m
@@ -612,6 +612,7 @@ static IOSClass *IOSClass_ArrayClassForName(NSString *name, NSUInteger index) {
 }
 
 IOSClass *IOSClass_forName_(NSString *className) {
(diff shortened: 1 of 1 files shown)


sfuku7:
Thank you for your comment !

Rollbacked.

tomball:
Looks good -- I'll pull this as soon as our internal repository is pushed, since commits in both repositories confuses our sync tool sometimes. :-)

tomball:
Thanks for your fix!

sfuku7:
Thanks for quick review and merge !

I’m happy I could contribute. 😄

PR #1006, opened 2018-10-09
Add missing inheritedChannel method

tux-mind:
Taken from https://android.googlesource.com/platform/libcore/+/nougat-release/ojluni/src/main/java/java/lang/System.java

### DIFF
--- a/jre_emul/Classes/java/lang/System.java
+++ b/jre_emul/Classes/java/lang/System.java
@@ -54,6 +54,8 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.nio.channels.Channel;
+import java.nio.channels.spi.SelectorProvider;
 
 /**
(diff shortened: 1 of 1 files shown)


tomball:
You know [this method always returns null on Android](https://android.googlesource.com/platform/libcore/+/master/ojluni/src/main/java/sun/nio/ch/PollSelectorProvider.java#40), right? I'm fine adding this if it makes porting some library easier, but there's no point if that library expects something non-null.

tux-mind:
I didn't check it, thank you for pointing it out :)

Docs are fine with null, libraries that use it shall handle the null case:
https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#inheritedChannel--

And yes, it will make the porting of external libraries easier!

I hope that this helps.
Thank you :)

Il giorno mar 9 ott 2018 alle ore 16:36 Tom Ball <notifications@github.com>
ha scritto:

PR #1017, opened 2018-12-07
Use correct hostname for wrapped ssl connections

FD-:
I've found another minor issue with the ssl socket implementation: For wrapped sockets, we are using the inetaddress of the underlying socket for SSLSetPeerDomainName. This is problematic, as with wrapped sockets, the underlying (wrapped) socket might be connected to a different host than the upper (wrapper) socket. An example for this would be an http proxy that is secured with ssl and is used to access an https server.
This commit fixes this problem in a way that is similar to how the [Conscrypt SSL provider](https://github.com/google/conscrypt/blob/8ea8a56875cfb109d52b436b8ebc0d1ca8b8e5f8/common/src/main/java/org/conscrypt/AbstractConscryptSocket.java) does it.


antonio-cortes-perez:
Thank you very much for the fix! I'll merge the pull request during our next repo sync.

PR #1044, opened 2019-05-11
new reserved names added

losharik:
When using the --swift-friendly flag, the currency is converted to property with the BSD name, but in iOS sdk there is already a define with the same name in header file param.h. There is also define EMPTY in the header utmpx.h.


losharik:
up

PR #1220, opened 2020-03-20
Fix jar template issue in xalan build

fluetm:
Fixes this problem when building j2objc on a Mac:

```
Building libxalan.a
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo -create /Users/matt.fluet/repo/cross-platform/j2objc/xalan/build_result/objs-appletvos/libxalan.a /Users/matt.fluet/repo/cross-platform/j2objc/xalan/build_result/objs-appletvsimulator/libxalan.a -output /Users/matt.fluet/repo/cross-platform/j2objc/xalan/build_result/appletvos/libxalan.a
building j2objc_xalan.jar
mktemp: too few X's in template ‘j2objc-xalan’
```

### DIFF
--- a/jre_emul/java.mk
+++ b/jre_emul/java.mk
@@ -27,6 +27,8 @@ ALL_JAVA_SOURCES = $(JAVA_SOURCES) $(NO_TRANSLATE_JAVA_SOURCES)
 
 ANNOTATIONS_JAR = $(DIST_JAR_DIR)/j2objc_annotations.jar
 
+MKTEMP_DIR = j2objc-jre_emul
+
 clean:
 	@rm -f $(EMULATION_JAR_DIST) $(EMULATION_SRC_JAR_DIST)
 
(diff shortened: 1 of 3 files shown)


fluetm:
Opened this against the wrong fork :octocat:

fluetm:
@googlebot I signed it!

fluetm:
@tomball having trouble doing a `make clean dist` from the tip of master on a Mac.  This seems to correct it for me...does this fix seem right?

tomball:
It looks like the correct answer is in [jre_emul/java.mk:61](https://github.com/google/j2objc/blob/master/jre_emul/java.mk#L61), which switches depending on what release of mktemp is used (thanks to @kstanger for that fix). Different mktemp's would explain why you're seeing this problem and we aren't.

So the quick fix is to copy those lines from java.mk into Xalan's Makefile, then use the MKTEMP_CMD variable in line 84. A better fix (avoids violating [the Don't Repeat Yourself principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)) is to move those lines from java.mk into make/common.mk, so both projects share the same build variable. That way, if a new implementation of mktemp is every made, the build breakage will be more obvious.

Want to make this change? If not, let me know and I'll take care of it. Either way, thanks for contributing to our project!

fluetm:
Sure...I can try out  what you're suggesting and verify that it solves the problem for us.  I'll push up an amended commit soon.  Thanks for the quick response!

fluetm:
@tomball how does this commit look?  Works locally on a Mac laptop

PR #1223, opened 2020-03-30
Enable --nullability by default (can be disabled with --no-nullability).

j2objc-copybara:
Enable --nullability by default (can be disabled with --no-nullability).


tomball:
Dropped because several projects reported build failures due to nullability warnings combined with -Werror.

Issue #1256, opened 2020-05-12
Generate ARC check in implementation files


tomball:
Although j2objc generates ARC or non-ARC source depending on its mutually-exclusive -use-arc flag or -use-reference-counting flags, compiling them with the wrong option (-fobjc-arc for ARC) may still succeed but not manage memory as expected.

To avoid this, implementation files should have a test near their beginning that verifies that they are compiled as their flags specified.

Default or -use-reference-counting:
```
#if __has_feature(objc_arc)
#error "<filename> should not be compiled with ARC (-fobjc-arc)"
#endif
```
use-arc:
```
#if !__has_feature(objc_arc)
#error "<filename> must be compiled with ARC (-fobjc-arc)"
#endif
```

albertomisail:
Should we close this? It was merged in 95783b5c87608a0aabc406075e7136d2b7f8a4ff

tomball:
Thanks for reminding me.

PR #1260, opened 2020-05-19
Enhancement #1256 Add generation of ARC check in implementation files

albertomisail:


tomball:
Merged using copybara.

PR #1267, opened 2020-05-26
changed time zone id from dawson to santa_isabel

anjulij:


tomball:
LGTM

anjulij:
@googlebot I signed it!

PR #1280, opened 2020-05-30
Adding tck test for java.time.temporal

albertomisail:
Test were added as is from libcore, then they were migrated from testng to junit


albertomisail:
Ok, I see your point. Just submitted the PR with the unchanged files (https://github.com/google/j2objc/pull/1281). Will leave this one open for now and sync once the other one gets updated.

PR #1300, opened 2020-06-04
TCK Chrono serial tests

albertomisail:


jre_emul/test_sources.mk
```diff
@@ -956,6 +956,10 @@ TEST_SOURCES := \
     org/apache/harmony/tests/org/xml/sax/helpers/XMLReaderFactoryTest.java \
     org/xmlpull/v1/XmlPullParserFactoryTest.java \
     sun/misc/UnsafeTest.java \
+    tck/java/time/chrono/serial/TCKChronoLocalDateTimeSerialization.java \
+    tck/java/time/chrono/serial/TCKChronoZonedDateTimeSerialization.java \
+    tck/java/time/chrono/serial/TCKChronologySerialization.java \
+    tck/java/time/chrono/serial/TCKEraSerialization.java \
```
tomball, line 962: Please copy these four entries, and add them to the TESTS_USE_SERIALIZATION section. In our internal build, we also build a jre_emul with reflection stripped, which causes all reflection and serialization tests to fail unless they are in these exclusion lists.

PR #1333, opened 2020-06-15
Updating javax.security.cert

albertomisail:


jre_emul/android/platform/libcore/ojluni/src/main/java/javax/security/cert/Certificate.java
```diff
@@ -1,79 +1,97 @@
...
-        }
-        if (!(obj instanceof Certificate)) {
+        if (!(other instanceof Certificate))
             return false;
-        }
-        Certificate object = (Certificate) obj;
+        Certificate otherCert = (Certificate) other;
         try {
-            return Arrays.equals(getEncoded(), object.getEncoded());
+            return Arrays.equals(getEncoded(), otherCert.getEncoded());
```
tomball, line 84: Add comment that this was changed for j2objc. We forgot to do so when we first made that fix, and I don't want it forgotten in future merges.

albertomisail: done

PR #1334, opened 2020-06-16
Update Arrays, ArrayIndexOutOfBoundsException to latest Android source, removing hidden API.

anjulij:
Updated a method in ArrayIndexOutOfBoundsException and consequently updated all references to said method

### DIFF
--- a/jre_emul/android/platform/libcore/ojluni/src/main/java/java/io/OutputStreamWriter.java
+++ b/jre_emul/android/platform/libcore/ojluni/src/main/java/java/io/OutputStreamWriter.java
@@ -261,7 +261,6 @@ public class OutputStreamWriter extends Writer {
     public void write(char[] buffer, int offset, int count) throws IOException {
         synchronized (lock) {
             checkStatus();
-            Arrays.checkOffsetAndCount(buffer.length, offset, count);
             CharBuffer chars = CharBuffer.wrap(buffer, offset, count);
             convert(chars);
         }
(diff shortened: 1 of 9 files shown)


jre_emul/android/platform/libcore/ojluni/src/main/java/java/io/OutputStreamWriter.java
```diff
@@ -308,17 +307,15 @@ public void write(int oneChar) throws IOException {
      *             {@code str}.
      */
     @Override
+    //TODO: use StreamEncoder per the update
```
tomball, line 310: In the future, TODO messages take the format "TODO(ldap): comment", so please add your Google ldap. There isn't a need to do so now, since you'll be updating this file soon.

jre_emul/android/platform/libcore/ojluni/src/main/java/java/io/OutputStreamWriter.java
```diff
@@ -308,17 +307,15 @@ public void write(int oneChar) throws IOException {
...
     public void write(String str, int offset, int count) throws IOException {
         synchronized (lock) {
-            if (count < 0) {
-                throw new StringIndexOutOfBoundsException(str, offset, count);
+            if ((offset | count) < 0 || offset > str.length() - count) {
+                throw new StringIndexOutOfBoundsException(str + "offset=" + offset + "formatElementIndex=" + count);
             }
             if (str == null) {
```
tomball, line 316: This test is redundant since the str.length() call above would through the NullPointerException, but there's no need to fix it here since it will be replaced by the StreamEncoder call.

jre_emul/android/platform/libcore/ojluni/src/main/java/java/io/OutputStreamWriter.java
anjulij: Will do. Thanks

jre_emul/android/platform/libcore/ojluni/src/main/java/java/io/OutputStreamWriter.java
anjulij: Woops, missed that. Thank you

PR #1354, opened 2020-06-25
Updating xcode proj and stop tracking workspace.xml

albertomisail:


tomball:
This PR failed importing by Copybara, probably because the workspace.xml file is in a directory listed in the project's .gitignore file.

PR #1371, opened 2020-07-09
Creating library for java.nio.file

albertomisail:
Initially sources of java.nio.file were going to be added to the channels library. Given the number of new sources and trying to keep each library as small as possible, a new library that contains the sources for java.nio.file is created.


jre_emul/tests.mk
```diff
@@ -335,6 +336,10 @@ $(TESTS_DIR)/core_plus_concurrent:
 	@mkdir -p $(@D)
 	$(J2OBJCC) -ljre_concurrent -ljre_util -o $@ -ObjC
 
+$(TESTS_DIR)/core_plus_file:
+	@mkdir -p $(@D)
+	$(J2OBJCC) -ljre_file -o $@ -ObjC
```
tomball, line 341: The file subset doesn't have any dependencies on other subsets? If that's true, awesome!

albertomisail: Not for now! I am still working on adding the missing sources (Unix related for example) and dependencies are probably going to appear.

tomball: Got it, thanks.

PR #1376, opened 2020-07-10
VMStack update

anjulij:


jre_emul/Classes/dalvik/system/VMStack.java
```diff
@@ -14,11 +16,101 @@
...
+   * @return the requested class loader, or {@code null} if this is the
+   *         bootstrap class loader.
+   * @deprecated Use {@code ClassLoader.getClassLoader(sun.reflect.Reflection.getCallerClass())}.
+   *         Note that that can return {@link BootClassLoader} on Android where the RI
+   *         would have returned null.
+   */
+  @UnsupportedAppUsage
+  @FastNative
+  @Deprecated
+  native public static ClassLoader getCallingClassLoader();
```
tomball, line 47: There needs to be a matching native code implementation for this to have a static modifier. However, since all classes are "boot loaded" on j2objc (since we cannot support loading external code), let's use the old non-native implementation from line 22 (just return null), moved here for the doc-comment and annotations.

jre_emul/Classes/dalvik/system/VMStack.java
```diff
@@ -14,11 +16,101 @@
...
+
+  /**
+   * Returns the class of the caller's caller.
+   *
+   * @return the requested class, or {@code null}.
+   * @deprecated Use {@link sun.reflect.Reflection#getCallerClass()}.
+   */
+  @Deprecated
+  public static Class<?> getStackClass1() {
```
tomball, line 56: This can be implemented using something like (not tested):
```
StackTraceElement[] stackTrace = getThreadStackTrace(Thread.currentThread());
return Class.forName(stackTrace[2].getClassName());
```
Embedded native code may be faster.

jre_emul/Classes/dalvik/system/VMStack.java
```diff
@@ -14,11 +16,101 @@
...
+  }
+
+  /**
+   * Returns the class of the caller's caller's caller.
+   *
+   * @return the requested class, or {@code null}.
+   */
+  @UnsupportedAppUsage
+  @FastNative
+  native public static Class<?> getStackClass2();
```
tomball, line 67: This can be implemented using something like (not tested):
```
StackTraceElement[] stackTrace = getThreadStackTrace(Thread.currentThread());
return Class.forName(stackTrace[3].getClassName());
```
Embedded native code may be faster.

jre_emul/Classes/dalvik/system/VMStack.java
```diff
@@ -14,11 +16,101 @@
...
+
+  /**
+   * Returns the first ClassLoader on the call stack that isn't the
+   * bootstrap class loader.
+   */
+  @FastNative
+  public native static ClassLoader getClosestUserClassLoader();
```
tomball, line 74: Just return null for the boot ClassLoader.

jre_emul/Classes/dalvik/system/AnnotatedStackTraceElement.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * A class encapsulating a StackTraceElement and lock state. This adds
+ * critical thread state to the standard stack trace information, which
+ * can be used to detect deadlocks at the Java level.
+ *
+ * @hide
+ */
+@libcore.api.CorePlatformApi
+public class AnnotatedStackTraceElement {
```
tomball, line 27: I don't see how we can implement this natively.

jre_emul/Classes/dalvik/system/VMStack.java
```diff
@@ -14,11 +16,101 @@
...
+   * Retrieves an annotated stack trace from the specified thread.
+   *
+   * @param t
+   *      thread of interest
+   * @return an array of annotated stack frames, or null if the thread
+   *      doesn't have a stack trace (e.g. because it exited)
+   */
+  @libcore.api.CorePlatformApi
+  @FastNative
+  native public static AnnotatedStackTraceElement[]
```
tomball, line 98: Comment out, can't implement (AFAIK).

jre_emul/Classes/dalvik/system/VMStack.java
```diff
@@ -14,11 +16,101 @@
...
+   * @param t
+   *      thread of interest
+   * @param stackTraceElements
+   *      preallocated array for use when only the top of stack is
+   *      desired. Unused elements will be filled with null values.
+   * @return the number of elements filled
+   */
+  @UnsupportedAppUsage
+  @FastNative
+  native public static int fillStackTraceElements(Thread t,
```
tomball, line 114: Implement with something like:
```
StackTraceElement[] stackTrace = t.getStackTrace();
for (int i = 0; i < stackTraceElements.length; i++) {
  stackTraceElements[i] = stackTrace[i];
}
return stackTraceElements.length;

File: jre_emul/android/platform/libcore/ojluni/src/main/java/sun/reflect/Reflection.java [hunk 8]
```diff
@@ -29,14 +29,21 @@
...
 
 public class Reflection {
 
-    public static void ensureMemberAccess(Class currentClass,
-                                          Class memberClass,
+   /*  public static Class<?> getCallerClass() {
+        // This method (getCallerClass()) constitutes another stack frame,
+        // so we need to call getStackClass2() rather than getStackClass1().
+        return VMStack.getStackClass2();
+    } */
```
tomball, line 43: Mary's PR (https://github.com/google/j2objc/pull/1380/files) updates two java.util.concurrent.atomic classes to call Reflection.getCallerClass(), which is why I sent the note about you two needing to coordinate with each other. Perhaps you could pair-program an implementation for this method? If I understand its requirements, this can be implemented with the following pseudo-code (not compiled or tested, so your solution may be different):
```
try {
    StackTraceElement[] stack = new Throwable().getStackTrace();
    return Class.forName(stack[2].getClassName());  // index may be wrong, verify with test
} catch (ClassNotFoundException e) {
    // shouldn't happen if the class name is right, but StackTraceElement.getClassName() might return a camel-cased name, null, or ???
} catch (???) {
}
```

PR #1381, opened 2020-07-14
Updating UnixNativeDispatcher, UnixFileAttributes and UnixFileStoreAttributes

albertomisail:


jre_emul/Classes/jni.m
```diff
@@ -455,6 +455,11 @@ static jint ThrowNew(JNIEnv *env, jclass clazz, const char *message) {
   return 0;
 }
 
+static jint EnsureLocalCapacity(JNIEnv *env, jint capacity) {
+  // no-op
+  return 0;
+}
+
```
tomball, line 462: Good call, we don't need it: https://stackoverflow.com/questions/7081853/jni-ensurelocalcapacity-why

jre_emul/jre_sources.mk
```diff
@@ -2438,5 +2444,5 @@ JAVA_PUBLIC_SOURCES = $(JAVA_PUBLIC_SOURCES_JRE) $(ANDROID_PUBLIC_SOURCES) \
 JAVA_SOURCES = $(JAVA_PUBLIC_SOURCES) $(JAVA_PRIVATE_SOURCES)
 
 NATIVE_JRE_SOURCES = $(NATIVE_JRE_SOURCES_CORE) $(NATIVE_JRE_SOURCES_ZIP) \
-  $(NATIVE_JRE_SOURCES_CHANNELS) $(NATIVE_JRE_SOURCES_NET) $(NATIVE_JRE_SOURCES_SSL) \
-  $(NATIVE_JRE_ICU_EMBEDDED_DATA)
+  $(NATIVE_JRE_SOURCES_CHANNELS) $(NATIVE_JRE_SOURCES_FILE) $(NATIVE_JRE_SOURCES_NET) $(NATIVE_JRE_SOURCES_SSL) \
+  $(NATIVE_JRE_ICU_EMBEDDED_DATA)
```
tomball, line 2448: Please move NATIVE_JRE_SOURCES_SSL to the next line (keeping the backslash on line 2447). We don't strictly enforce line lengths on make files (some paths are just too long), but really long lines are harder to read in some IDEs.

jre_emul/Classes/jni.m
albertomisail: Yes. That is why I left is a no-op for now. Nonetheless, the declaration of the method is needed in jni_util.m:110 in JNU_NewObjectByName

jre_emul/jre_sources.mk
albertomisail: Done

PR #1388, opened 2020-07-17
update Reflection.getCallerClass

maryqyh:
added Reflection.getCallerClass and uncommented related methods

### DIFF
--- a/jre_emul/android/platform/libcore/ojluni/src/main/java/java/util/concurrent/atomic/AtomicIntegerFieldUpdater.java
+++ b/jre_emul/android/platform/libcore/ojluni/src/main/java/java/util/concurrent/atomic/AtomicIntegerFieldUpdater.java
@@ -40,6 +40,7 @@ import java.lang.reflect.Modifier;
 import java.util.function.IntBinaryOperator;
 import java.util.function.IntUnaryOperator;
 import sun.reflect.CallerSensitive;
+import sun.reflect.Reflection;
 
 /**
  * A reflection-based utility that enables atomic updates to
@@ -79,11 +80,8 @@ public abstract class AtomicIntegerFieldUpdater<T> {
     @CallerSensitive
     public static <U> AtomicIntegerFieldUpdater<U> newUpdater(Class<U> tclass,
                                                               String fieldName) {
-        return new AtomicIntegerFieldUpdaterImpl<U>(tclass, fieldName, null);
-        /* J2ObjC: Call stack not available.
         return new AtomicIntegerFieldUpdaterImpl<U>
-            (tclass, fieldName, Reflection.getCallerClass()); // android-changed
-        */
+            (tclass, fieldName, Reflection.getCallerClass());
     }
 
     /**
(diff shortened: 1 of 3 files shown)


jre_emul/android/platform/libcore/ojluni/src/main/java/java/util/concurrent/atomic/AtomicIntegerFieldUpdater.java
```diff
@@ -79,11 +80,8 @@
     @CallerSensitive
     public static <U> AtomicIntegerFieldUpdater<U> newUpdater(Class<U> tclass,
                                                               String fieldName) {
-        return new AtomicIntegerFieldUpdaterImpl<U>(tclass, fieldName, null);
-        /* J2ObjC: Call stack not available.
         return new AtomicIntegerFieldUpdaterImpl<U>
-            (tclass, fieldName, Reflection.getCallerClass()); // android-changed
-        */
+            (tclass, fieldName, Reflection.getCallerClass());
```
tomball, line 84: Nit: restore the "// android-changed" comment from the libcore source.

jre_emul/android/platform/libcore/ojluni/src/main/java/java/util/concurrent/atomic/AtomicReferenceFieldUpdater.java
```diff
@@ -101,11 +102,8 @@
     public static <U,W> AtomicReferenceFieldUpdater<U,W> newUpdater(Class<U> tclass,
                                                                     Class<W> vclass,
                                                                     String fieldName) {
-        return new AtomicReferenceFieldUpdaterImpl<U,W>(tclass, vclass, fieldName, null);
-        /* J2ObjC: Call stack not available.
         return new AtomicReferenceFieldUpdaterImpl<U,W>
-            (tclass, vclass, fieldName, Reflection.getCallerClass()); // android-changed
-        */
+            (tclass, vclass, fieldName, Reflection.getCallerClass());
```
tomball, line 106: Ditto: restore libcore comment.

jre_emul/android/platform/libcore/ojluni/src/main/java/java/util/concurrent/atomic/AtomicIntegerFieldUpdater.java
maryqyh: The "// android-changed" comment is actually only in libcore-nougat but not libcore, so I delete it in this commit

jre_emul/android/platform/libcore/ojluni/src/main/java/java/util/concurrent/atomic/AtomicReferenceFieldUpdater.java
maryqyh: The "// android-changed" comment is only in libcore-nougat but not libcore

jre_emul/android/platform/libcore/ojluni/src/main/java/java/util/concurrent/atomic/AtomicIntegerFieldUpdater.java
tomball: Perfect, thanks for the update.

tomball:
Please change Reflection.getCallerClass() to:
```
    public static Class<?> getCallerClass() {
        return VMStack.getStackClass2();
    }
```
Anjuli separately submitted the getStackClass2() update that uses a Throwable's stack frames like you did. However, we didn't skip the three stack frames from Throwable, so instead of `stack[2]` it should be `stack[5]`. I updated that in a separate CL which should be merged later this afternoon. With your change, these tests will now pass.

PR #1393, opened 2020-07-20
update util.zip.testZipFileOffsetNeverChangesAfterInit

maryqyh:
fixed OsConstants and offset issues in util.zip.testZipFileOffsetNeverChangesAfterInit

### DIFF
--- a/jre_emul/android/platform/libcore/luni/src/test/java/libcore/java/util/zip/ZipFileTest.java
+++ b/jre_emul/android/platform/libcore/luni/src/test/java/libcore/java/util/zip/ZipFileTest.java
@@ -18,7 +18,7 @@ package libcore.java.util.zip;
 
 /* J2ObjC modified.
 import android.system.OsConstants; */
-import static libcore.io.OsConstants.*;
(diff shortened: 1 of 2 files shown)


jre_emul/android/platform/libcore/luni/src/test/java/libcore/java/util/zip/ZipFileTest.java
```diff
@@ -16,8 +16,9 @@
 
 package libcore.java.util.zip;
 
-/* J2ObjC removed.
+/* J2ObjC modified.
 import android.system.OsConstants; */
+import static libcore.io.OsConstants.*;
```
lukhnos, line 21: Nit: since we only need `Libcore.os.lseek` and `Libcore.os.SEEK_CUR`, it may be better we just static-import those two symbols, instead of importing all with `*`.

maryqyh: changed the static import for SEEK_CUR. I think lseek() requires `import libcore.io.Libcore;`

PR #1400, opened 2020-07-24
fix getCallerClass in concurrent.atomic

maryqyh:
commented out the use of getCallerClass in AtomicReferenceFieldUpdater, because the returned caller class name was only used for checkPackageAccess, while checkPackageAccess was removed in latest update due to "Skip checkPackageAccess which is a noop on Android."


tomball:
I fixed this in PR #1402, as the Calendar team reported it as a performance regression. I'm sorry for the duplicated effort!

PR #1405, opened 2020-07-25
Finished refactoring translator for preparing to merge into master.

slowcoders:
Difference with J2ObjC2.6 master.

1) Option added
    --use-gc // use garbage collection.
    --not-import // specifies not importable class or packages
    -Xiostest // convert junit testcase into IOS testcase(XCTestCase)
    -Xconst-ref-args // enables method __attribute__((objc_externally_retained)) for reference-count performance optimization 
2) Added class
    SourceStore // InputSource cache
    ImportManager, FileManagerProxy, NotImportedError // for implement --not-import option


tomball:
This overall looks great, just a few questions and comments:

Please change `--use-gc` to `-Xuse-gc`, so that we can migrate to GC while indicating it's experimental (changing back to `-use-gc` should be the last step).

`--not-import` is an awkward name (since there's no related `--import`). Maybe change to `--ignore` or `--skip`? Let's discuss.

Does `-Xiostest` work for JUnit 4 or just JUnit 3? Does it work with [@RunWith](https://junit.org/junit4/javadoc/latest/org/junit/runner/RunWith.html)? It doesn't have to (that's hard), but an error should be reported for any unsupported JUnit test situation when this flag is set.

Will using `-Xconst-ref-args` fail if `--use-gc` is not specified? One option to have several flags (and checking that they are only used for the right mode) is to add a colon to support sub-options. For example, to support multiple source-level debugging modes, we have `-g` (Java source debugging, full paths), `-g:none` (no Java source debugging, to allow ObjC source debugging), and `-g:relative` (Java source debugging, relative paths). So maybe `--use-gc:const-ref-args`?

Do you know of any performance issues with the new SourceStore, ImportManager, or FileManagerProxy? Build speed is one of the top priorities for Google's iOS app teams.

slowcoders:
Dear Tom Ball.

1) --not-import
Naming is the most difficult thing. ^^;;  How about --exclude?

2) -Xiostest
It does not support @RunWith. It is a very simple implementation that just
changes the super class to the XCTestCase.
XCTestCase can be executed in Xcode UI (= like JUnit runs in Eclipse)

3) -Xconst-ref-args
It does not depend on ARGC. I fixed bugs. (It will be committed tomorrow)

4) Translator optimization.
SourceStore caches all the input sources. It reduces source search time
especially sources in JarFile.
ImportManager and FileManagerProxy are utility classes for implementing the
--not-import feature. Those are used only with --not-import



2020년 7월 28일 (화) 오전 3:01, Tom Ball <notifications@github.com>님이 작성:

tomball:
On Tue, Jul 28, 2020 at 5:27 AM Daehoon Zee <notifications@github.com>
wrote:

> Dear Tom Ball.
>
> 1) --not-import
> Naming is the most difficult thing. ^^;; How about --exclude?
>

--exclude sounds good.


> 2) -Xiostest
> It does not support @RunWith. It is a very simple implementation that just
> changes the super class to the XCTestCase.
> XCTestCase can be executed in Xcode UI (= like JUnit runs in Eclipse)
>

SGTM (sounds good to me).

3) -Xconst-ref-args
> It does not depend on ARGC. I fixed bugs. (It will be committed tomorrow)
>

SGTM

4) Translator optimization.
> SourceStore caches all the input sources. It reduces source search time
> especially sources in JarFile.
> ImportManager and FileManagerProxy are utility classes for implementing the
> --not-import feature. Those are used only with --not-import
>

SGTM, thanks for the answers.

slowcoders:
Dear Tom Ball.

I have finished refactoring the ARGC branch to prepare to merge into Master.
I fully passed the default J2ObjC build and tests.
I'd appreciate it if you could refine and rearrange source codes.

And, let me ask a question.
Do you mind if I am going to change the translator to generate a single
source that can be built with any memory model options?
I think it is possible to build J2ObjC with arc, reference-counting and
ARGC version with the same source files.


2020년 8월 1일 (토) 오전 4:21, Tom Ball <notifications@github.com>님이 작성:

tomball:
Is it reasonable for us to cherry-pick the bug fixes first which are unrelated to ARGC? This is a huge delta, but several of your changes are fixes that should be accepted independently. Several of these were "retain -> RETAIN_", "release -> RELEASE", etc. changes, while others (such as in IosRsaKey) fixed memory leaks.

After the bug fixes, I'd like to pull in the new features that are not ARGC related, each in a separate PR: string interning, --exclude, class initialization changes, etc. Having each in a separate PR allows them to be easily rolled back if any internal apps break (we have strict policies about keeping the shared repository passing all tests). When only ARGC-specific commits are left, that PR will be much more easily streamlined into our main base.

I haven't cherry-picked commits using Git before, but am happy to do this if you don't the time or interest.

slowcoders:
1) About cherry-picked commits.
Yes. I hope you're working on that.
Please feel free to refine the naming and architectures more naturally.
I will prepare the next version of garbage collection.

2020년 8월 5일 (수) 오전 2:00, Tom Ball <notifications@github.com>님이 작성:

PR #1408, opened 2020-07-29
Updating java.nio.file and java.nio.channels

albertomisail:


jre_emul/android/platform/libcore/ojluni/src/main/java/sun/nio/ch/AsynchronousChannelGroupImpl.java
```diff
@@ -62,6 +64,7 @@
 
     // task queue for when using a fixed thread pool. In that case, thread
     // waiting on I/O events must be awokon to poll tasks from this queue.
+    @Weak
     private final Queue<Runnable> taskQueue;
```
tomball, line 68: This looks wrong, as the taskQueue is created in the constructor and never made visible to a caller, so it needs to be a strong reference (the default). 

What does the cycle error look like?

albertomisail: done

PR #1411, opened 2020-07-30
Updating sun.nio.ch

albertomisail:
These sources are required in order to add DefaultAsynchronousChannelProvider (used in most tests of java.nio.channels) in the future


tomball:
I had to remove the jre_concurrent dependency from tests.mk:core_plus_channels because several internal tests had the same separation of subsets. Since there were several unresolved links from that, I added factory methods for these types to sun/nio/ch/Util.java, so all the reflection code is in one place.

albertomisail:
SGTM

PR #1415, opened 2020-08-03
update java.net

maryqyh:


jre_emul/android/platform/libcore/ojluni/src/main/java/java/net/Inet6Address.java
```diff
@@ -138,138 +142,320 @@
...
+
+        // NetworkInterface has two strong references to InetAddress:
+        //  1. NetworkInterface.addrs[] -> InetAddress
+        //  2. NetworkInterface.bindings[] -> InterfaceAddress.address -> InetAddress.
+        // And this class which is a subclass of InetAddress has a field of type NetworkInterface. To
+        // resolve this reference cycle, we can either use the RETAINED_WITH_CHILD_NUM_REFS macro and
+        // set NUM_REFS to 2 or cache all the network interfaces in a map with String keys since the
+        // number of network interfaces on an iOS device is small.
+        /*-[
+        RETAINED_WITH_CHILD_NUM_REFS(scope_ifname_, 2)
```
maryqyh, line 255: The macro is moved from an outer class field to an inner class field, thus may need to test for potential retain cycle.

tomball: You should test for cycles by running "make -f java.mk find_cycles".

PR #1423, opened 2020-08-06
update DeleteOnExitHook.java

anjulij:
- replaced luni/src/main/java/libcore/io/DeleteOnExit.java with ojluni/src/main/java/java/io/DeleteOnExitHook.java 

- Modified usage of DeleteOnExit to DeleteOnExitHook in io/File.java


jre_emul/android/platform/libcore/ojluni/src/main/java/java/io/File.java
```diff
@@ -43,12 +43,14 @@
...
 import java.net.URL;
+import java.nio.file.FileSystems;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 import java.util.Random;
 
-import libcore.io.DeleteOnExit;
+//import libcore.io.DeleteOnExitHook;
```
tomball, line 53: Please delete line.

PR #1428, opened 2020-08-07
Adding method toPath to io.File

albertomisail:


jre_emul/android/platform/libcore/ojluni/src/main/java/java/io/File.java
```diff
@@ -1281,4 +1284,73 @@ private static native boolean isCaseSensitiveImpl() /*-[
...
+                result = filePath;
+                if (result == null) {
+                    try {
+                        Method getDefaultMethod;
+                        try {
+                            getDefaultMethod = Class.forName("java.nio.file.FileSystems")
+                                    .getDeclaredMethod("getDefault");  // no parameters
+                        } catch (ClassNotFoundException e) {
+                            throw new LibraryNotLinkedError("Get default FileSystem",
+                                    "jre_file", "java.nio.file.FileSystems");
```
tomball, line 1331: The first parameter should say what library subset missing, so "java.nio.file support" is more general. 

The last parameter needs to be camel-cased (so "JavaNioFileFileSystems"). This is because that's the missing symbol name which gets reported as a linker error.

jre_emul/android/platform/libcore/ojluni/src/main/java/java/io/File.java
```diff
@@ -1281,4 +1284,73 @@ private static native boolean isCaseSensitiveImpl() /*-[
...
+
+                        Object fileSystem = getDefaultMethod.invoke(null);    // null, since it's a static method.
+
+                        Method getPathMethod;
+                        try {
+                            getPathMethod = Class.forName("java.nio.file.FileSystem")
+                                    .getDeclaredMethod("getPath", String.class);
+                        } catch (ClassNotFoundException e) {
+                            throw new LibraryNotLinkedError("Get path of FileSystem",
+                                    "jre_file", "java.nio.file.FileSystem");
```
tomball, line 1342: Same comment, with the different symbol name.

albertomisail: Fixed

PR #1429, opened 2020-08-07
Updating IO

anjulij:
- Updated comments and copyright
- modification to lineSeparator assignment
- Use of current instance instead of WeakProxy
- modification of close() in FilterOutputStream

### DIFF
--- a/jre_emul/android/platform/libcore/ojluni/src/main/java/java/io/PrintWriter.java
+++ b/jre_emul/android/platform/libcore/ojluni/src/main/java/java/io/PrintWriter.java
@@ -25,7 +25,6 @@
 
 package java.io;
 
-import com.google.j2objc.WeakProxy;
 import java.util.Objects;
 import java.util.Formatter;
 import java.util.Locale;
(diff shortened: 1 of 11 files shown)


jre_emul/android/platform/libcore/ojluni/src/main/java/java/io/PrintWriter.java
```diff
@@ -901,7 +901,7 @@ public PrintWriter format(String format, Object ... args) {
                 ensureOpen();
                 if ((formatter == null)
                     || (formatter.locale() != Locale.getDefault()))
-                    formatter = new Formatter(WeakProxy.forObject(this));
+                    formatter = new Formatter(this);
```
tomball, line 904: Please revert, as `new Formatter(this)` causes a reference cycle which WeakProxy avoids.

jre_emul/android/platform/libcore/ojluni/src/main/java/java/io/PrintWriter.java
```diff
@@ -960,7 +960,7 @@ public PrintWriter format(Locale l, String format, Object ... args) {
             synchronized (lock) {
                 ensureOpen();
                 if ((formatter == null) || (formatter.locale() != l))
-                    formatter = new Formatter(WeakProxy.forObject(this), l);
+                    formatter = new Formatter(this, l);
```
tomball, line 963: Revert (see above).

PR #1438, opened 2020-08-12
Adding test.java.nio.file.attribute

albertomisail:


jre_emul/android/platform/libcore/ojluni/src/main/java/sun/nio/fs/UnixFileModeAttribute.java
```diff
@@ -47,7 +47,11 @@ private UnixFileModeAttribute() {
 
     static int toUnixMode(Set<PosixFilePermission> perms) {
         int mode = 0;
-        for (PosixFilePermission perm: perms) {
+        /* J2ObjC changed: explicitly cast as if an element of perms is not
+        of type PosixFilePermission an ClassCastException is not thrown
+         */
```
tomball, line 52: Nits: "as if" => "in case that", "an ClassCastException is not" => ", so that a ClassCastException is".

albertomisail: done

PR #1444, opened 2020-08-13
update java.util

maryqyh:


jre_emul/android/platform/libcore/ojluni/src/main/java/java/util/ArrayList.java
```diff
@@ -1473,19 +1473,4 @@ public void sort(Comparator<? super E> c) {
...
-      if (state->state == 0) {
-        state->mutationsPtr = (unsigned long *) &modCount_;
-        state->itemsPtr = (__unsafe_unretained id *) (void *) elementData_->buffer_;
-        state->state = 1;
-        return size_;
-      } else {
-        return 0;
-      }
-    }
-    ]-*/
```
tomball, line 1490: This needs to be restored -- it supports Objective C fast enumeration.

maryqyh: Thanks for pointing out! Restored the change. I ran the cycle-finder test and it said "0 CYCLES FOUND."

tomball: Perfect, thanks!

PR #1517, opened 2020-11-22
Add support for the arm64 architecture for macOS

dsanghan:
Verified with:

```
> lipo -archs jre_emul/build_result/macosx/libjre_emul.a 
x86_64 arm64
```

Is anything else needed?

### DIFF
--- a/make/common.mk
+++ b/make/common.mk
@@ -76,7 +76,7 @@ J2OBJC_ARCHS = $(subst _, ,$(ENV_J2OBJC_ARCHS))
 else
 # 32bit iPhone archs are no longer built by default. To build a release
 # with them, define J2OBJC_ARCHS with "iphone" and "simulator" included.
-J2OBJC_ARCHS = macosx iphone64 iphone64e watchv7k watch64 watchsimulator \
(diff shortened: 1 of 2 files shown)


dsanghan:
`make dist` completes correctly, but `make frameworks` outputs an error:
`Both macos-arm64 and macos-x86_64 represent two equivalent library definitions.`

From the forums it seems we're supposed to use lipo for this - so use build_result/macosx/lib instead of build_results/objs_/lib - but not sure how to do that

dsanghan:
That does seem to work - making the above change does generate a xcframework: 

Screenshot 2020-11-22 at 5 33 39 PM

tomball:
Unfortunately `xcodebuild -create-framework` fails because it considers `macos-x86_64` and `macos-arm64` to "represent two equivalent definitions." [This Apple Developer Forum discussion has the fix.](https://developer.apple.com/forums/thread/666335)

Adding a `lipo` step to some of the framework slices involves more Make hackery than just adding a new architecture. Do you want to give it a shot, or should I add it in a follow-up commit?

dsanghan:
Would be great if you could add a follow-up commit

tomball (approved):
LGTM -- thanks!

PR #1646, opened 2021-04-04
Removed redundant library builds, updated distribution script, ignored generated Python scripts.

tomball:
Removed redundant library builds, updated distribution script, ignored generated Python scripts.

GOOGLE: N/A


dinooliva (approved):
lgtm

PR #1650, opened 2021-04-14
Add support for the arm64 architecture for Mac Catalyst

albouan:


tomball (commented):
Looks good, and you get bonus points for being the first external contributor brave enough to tackle the make macros. :-)

The maccatalyst64 arch should be optionally included in [J2OBJC_ARCHS](https://github.com/google/j2objc/blob/master/make/common.mk#L90) in make/common.mk, when the xcodebuild version supports building that arch. 

Please look at the [MACOSX64_AVAILABLE](https://github.com/google/j2objc/blob/master/make/common.mk#L76) definition: is the maccatalyst64 arch supported in the same Xcode version as macosx64? If so, then maccatalyst64 can be added to [line 93](https://github.com/google/j2objc/blob/master/make/common.mk#L93). If not, then copy and paste MACOSX64_AVAILABLE to a MACCATALYST64_AVAILABLE macro, and copy/paste lines 92-94.

albouan:
There is no straightforward way to detect arm64 Mac Catalyst availability using xcodebuild. Xcode 12 and later seemed reasonable and was used as the inclusion criteria since universal macOS binaries was introduced with it.

tomball (approved):
Looks good, thanks.

PR #1672, opened 2021-05-06
Make the ARGC branch buildable

lukhnos:
This fixes a few accidental imports that should have been from java.io.

Tested: `make dist` now works.

### DIFF
--- a/jre_emul/android/platform/libcore/ojluni/src/main/java/java/lang/reflect/Proxy.java
+++ b/jre_emul/android/platform/libcore/ojluni/src/main/java/java/lang/reflect/Proxy.java
@@ -872,7 +872,10 @@ public class Proxy implements java.io.Serializable {
     ]-*/;
 
     native int proxy_hashCode() /*-[
+#pragma clang diagnostic push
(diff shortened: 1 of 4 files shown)


tomball (commented):
Would it be much more work if the new warning suppressions were moved into the sources that need them? [Here's an example.](https://github.com/google/j2objc/blob/master/jre_emul/Classes/java/lang/AbstractStringBuilder.m#L731)

lukhnos:
> Would it be much more work if the new warning suppressions were moved into the sources that need them? [Here's an example.](https://github.com/google/j2objc/blob/master/jre_emul/Classes/java/lang/AbstractStringBuilder.m#L731)

Done. Turns out that I should cherry-pick a commit from the mainline for the unsequenced expression warning (and so I did here now). The pointer-to-integer cast warning starts to be a thing as I've upgraded to Xcode 12.5, and the same warning is showing up in mainline too.

Which means this PR really just fixes some accidental imports in one file, and that makes the ARGC branch buildable again.

tomball (approved):
Thanks!

Issue #1737, opened 2021-08-03
Missing Arm64 Architecture for iOS Simulator


L6Echo:
For iOS Simulator the Arm64 Architecture is missing. This is required to run iOS Apps in Simulator on Macs with M1 chips.

tomball:
This should have been fixed in [821ff4e](https://github.com/google/j2objc/commit/821ff4ebe99ed79011b31aae73158b3f8b47b072). Since this pull request was merged after the last release, you'll need to [build from source](https://developers.google.com/j2objc/guides/building-j2objc). "Real" developers should normally be doing that anyway, since the releases are really just for people who want to experiment with this project before seriously using it.

L6Echo:
> "Real" developers should normally be doing that anyway, since the releases are really just for people who want to experiment with this project before seriously using it.

Thanks for the clarification. This was not clear to us. I can't find any comments, that these releases are only intended to be used for experimental testing. Also the "Getting Started" section says that this one way to go without limiting it to experimental tests.

> First, either:
> Download the current distribution from the Releases section and unzip it, or
> Get the source and build it.

tomball:
Right, it wasn't intended that way when we started (and wrote those docs).
What happened is that the big app teams kept wanting the latest (often the
moment it was committed), and gradually the only feedback on releases came
from new (to the project) developers.

BTW, I didn't mean my comment as a criticism in any way. Some teams have a
"use only released libraries" policy, which makes sense when they come from
teams with lots of QA and release engineers. Many people assume that j2objc
is an officially supported Google project, but it's just a small,
unofficial part of the Gmail team. Consequently, we early on adopted
test-driven development so that we could ensure high-quality with every
commit. So whenever a release is made, within days the repository HEAD has
a better version.

Issue #1736, opened 2021-08-03
Missing Simulator Architectures for WatchOS 6 and later


L6Echo:
For WatchOS 6 and later the Architecture x86_64 for Intel and arm64 for M1 is required. Currently only the Architecture i386, which was used for WatchOS 5 and earlier, is included in J2Objc watch lib.

tomball:
This should have been fixed in [821ff4e](https://github.com/google/j2objc/commit/821ff4ebe99ed79011b31aae73158b3f8b47b072). Since this pull request was merged after the last release, you'll need to [build from source](https://developers.google.com/j2objc/guides/building-j2objc). "Real" developers should normally be doing that anyway, since the releases are really just for people who want to experiment with this project before seriously using it.

tomball:
Please note: our team doesn't have any M1 hardware, since it's scarce and other Google iOS teams (with shipping apps) have higher priority.  If architectural slices are missing from j2objc's build, we're therefore relying on pull requests from external developers to add that support, like [821ff4e](https://github.com/google/j2objc/commit/821ff4ebe99ed79011b31aae73158b3f8b47b072) was.

L6Echo:
The mentioned commit adds "only" the arm64 for iPhone Simulator.

For watchOS Simulator only the architecture i386 can be build from source. i386 is only compatible for WatchOS 5 and earlier. When addressing watchOS 6 and later it is not possible right now to build for the Simulator (neither M1 nor Intel) neither with the release builds nor when compiling the source code.

I am not too deep into how the build scripts work, but I will try to add it and create a PR.

Issue #1713, opened 2021-06-16
Framework build not work with archs "iphone64 simulator64 iphone simulator"


sarsonj:
The similar bug #1567 was already there, but it looks like it appears again - at least with latest stable xCode (version 12.5). Tested with fresh git pull.

Result is still same: 

`export J2OBJC_ARCHS="iphone64 simulator64 iphone simulator"`
`make -j8 all_dist`

result:
```
building JRE framework
xcodebuild -create-xcframework -output ../dist/frameworks/JRE.xcframework -library /Users/sarsonj/work/git/j2objc/jre_emul/build_result/objs-iphone64/libjre_emul.a -library /Users/sarsonj/work/git/j2objc/jre_emul/build_result/objs-iphone/libjre_emul.a -library /Users/sarsonj/work/git/j2objc/jre_emul/build_result/objs-iphone64/libjre_emul.a
Both ios-armv7 and ios-arm64 represent two equivalent library definitions.
make[1]: *** [../dist/frameworks/JRE.xcframework] Error 70
make: *** [frameworks] Error 2
```

I tried to update framework.mk to ad `iphone` into `FMWK_FAT_LIBS` (according to comment in makefile, there should be there anyway?), but then `gen_xcframework.sh` failed with empty parameters, so that this simple fix didn't helps.

L6Echo:
The issue seems to be how the "fat libs" paths are constructed in framwork.mk:100

```
 94: # As of Xcode 12, xcframeworks need single slice libraries, except for iOS, simulator and Mac platforms
 95: # which require fat libraries.
 96: FMWK_ARCH_LIBS = $(filter-out simulator% macos% maccatalyst%,$(J2OBJC_ARCHS))
 97: FMWK_FAT_LIBS = $(filter simulator% macos% maccatalyst%,$(J2OBJC_ARCHS))
 98: framework_libraries = \
 99:  $(foreach arch,$(FMWK_ARCH_LIBS),$(wildcard $(BUILD_DIR)/objs-$(arch)*/lib$(1).a)) \
100:  $(foreach arch,$(FMWK_FAT_LIBS),$(wildcard $(BUILD_DIR)/$(arch)/lib$(1).a))
```

**General issue on constructing fat libraries:**

[...]

PR #1743, opened 2021-08-10
Fix construction of xcframework architectures

L6Echo:
This PR addresses some issues during the construction of the xcframework.

**Thin vs fat libraries**
When building the xcframework for each platform an own fat library has to be used. But for iphone, simulator, watchos, watchsimulator for each slice a thin library was used. That caused `xcodebuild -create-xcframework` complains with 
'Both x and y represent two equivalent library definitions.'

There was a workaround in `gen_xcframework` which just skipped the archs `iphone64e` and `watchv7k` and so they are also missing in the xcarchive. `watchv7k`is still required to upload a watchos app to app store.

**Missing archs**
For WatchOS the arches `x86_64` and `arm64` are not built. These are required to be able to run a WatchOS 6 or later app in a simulator. So current app watch apps are only able to be developed against a real device.

**Dependent changes**

**Fat library for iPhone moved to `dist/iphone`**
Currently all platforms has a own subfolder in `dist/[platform]` except for iPhone. It seems to me, that the intention was to have it too, because it was already prepared on same places. I changed that, because otherwise there are always exceptions needed. (e.g. finding the fat platform library for the xcframework)

**Fat library for iPhone/Simulator and WatchOs/Watchsimulator separated**
Before the fat library `dist/lib/libjre_emul.a` and `dist/lib/watchos/libjre_emul` contained architecture for the device and the simulator together. But because of introducing the arm64 simulator arches, they have to be separated. This was the reason why apple introduced the xcframeworks as well.

Together with the change before the new structure is:

- dist
  - lib
    - iphone
       - libjre_emul.a
    - simulator
       - libjre_emul.a
    - watchos
       - libjre_emul.a
    - watchsimulator 
       - libjre_emul.a

This has to be considered when linking the static libraries.

**Renamed watch architecture in `J2OBJC_ARCHS`**
Before `watchv7k` and `watch64` was used to configure the watchos targets via `J2OBJC_ARCHS`.
These were renamed to `watchosv7k` and `watchos64`. Otherwise it's hard to distinguish between `watch%` wand `watchsimulator%`.

The official architecture name is `watchos` anyway. And both notations was  inside the build scripts.

#1713 #1736 #1737

### DIFF
--- a/make/common.mk
+++ b/make/common.mk
@@ -57,6 +57,11 @@ ARCH_BUILD_WATCH_DIR = $(ARCH_BUILD_DIR)/watchos
 ARCH_LIB_WATCH_DIR = $(ARCH_LIB_DIR)/watchos
 DIST_LIB_WATCH_DIR = $(DIST_LIB_DIR)/watchos
 
+# AppleWatch simulator library dirs.
+ARCH_BUILD_WATCHSIMULATOR_DIR = $(ARCH_BUILD_DIR)/watchsimulator
+ARCH_LIB_WATCHSIMULATOR_DIR = $(ARCH_LIB_DIR)/watchsimulator
+DIST_LIB_WATCHSIMULATOR_DIR = $(DIST_LIB_DIR)/watchsimulator
+
 # Appletv library dirs.
 ARCH_BUILD_TV_DIR = $(ARCH_BUILD_DIR)/appletvos
 ARCH_LIB_TV_DIR = $(ARCH_LIB_DIR)/appletvos
--- a/make/common.mk
+++ b/make/common.mk
@@ -89,7 +89,7 @@ J2OBJC_ARCHS = $(subst _, ,$(ENV_J2OBJC_ARCHS))
 else
 # 32bit iPhone archs are no longer built by default. To build a release
 # with them, define J2OBJC_ARCHS with "iphone" and "simulator" included.
-J2OBJC_ARCHS = macosx iphone64 iphone64e watchv7k watch64 watchsimulator \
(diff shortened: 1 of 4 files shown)


L6Echo:
@googlebot I signed it!

tomball:
These changes look good (thanks!), but they can't be pulled until the CLA issue is resolved. Please read the "What to do if you already signed the CLA" comment above, and see if you can resolve the issue. If not, please email me the details at tball at google.com and I'll see if someone who supports our GitHub account can help.

L6Echo:
@googlebot I signed it!

L6Echo:
> These changes look good (thanks!), but they can't be pulled until the CLA issue is resolved. Please read the "What to do if you already signed the CLA" comment above, and see if you can resolve the issue. If not, please email me the details at tball at google.com and I'll see if someone who supports our GitHub account can help.

Sure. The corporate signing process was still in progress and is successfully finished now.

tomball (approved):
Thanks for your contribution!

tomball (approved):
Reapproving, since the previous attempt didn't trigger a repository sync.

L6Echo:
@tomball There some internal checks failed, but I can't see the issues. Do you have some hints for me? Then I will try to resolve it.

tomball:
We use [Copybara](https://github.com/google/copybara) to sync our public repository with our internal one, and the process to sync this pull request with our internal repository kicks off when I approved it last. Normally that works fine, but this time it hit a Copybara bug where the changed files in this PR were copied to new paths instead of updating the files, and those new paths broke the build. I filed a high priority bug with the Copybara team to get this fixed, and will try again as soon as that's done.

tomball:
Thanks for this contribution, and thanks for your patience while our repository sync tools were repaired!

Issue #2165, opened 2023-09-11
The 'build_subset_frameworks.sh' script fails


adil-hussain-84:
The [build_subset_frameworks.sh][4] script fails for me when I run it on a macOS Ventura 13.5.2 machine. I am running the following commands one by one:

```shell
export JAVA_HOME=$(/usr/libexec/java_home -v 11)
export J2OBJC_ARCHS="iphone64 simulator64 simulator"
make -j8 dist
cd jre_emul
./build_subset_frameworks.sh
```

The final command in the sequence above produces the following output:

```none
make: *** No rule to make target `lib', needed by `framework'.  Stop.
make: *** No rule to make target `lib', needed by `framework'.  Stop.
make: *** No rule to make target `lib', needed by `framework'.  Stop.
make: *** No rule to make target `lib', needed by `framework'.  Stop.
make: *** No rule to make target `lib', needed by `framework'.  Stop.
make: *** No rule to make target `lib', needed by `framework'.  Stop.
make: *** No rule to make target `lib', needed by `framework'.  Stop.
make: *** No rule to make target `lib', needed by `framework'.  Stop.
make: *** No rule to make target `lib', needed by `framework'.  Stop.
make: *** No rule to make target `lib', needed by `framework'.  Stop.
make: *** No rule to make target `lib', needed by `framework'.  Stop.
make: *** No rule to make target `lib', needed by `framework'.  Stop.
make: *** No rule to make target `lib', needed by `framework'.  Stop.
make: *** No rule to make target `lib', needed by `framework'.  Stop.
make: *** No rule to make target `lib', needed by `framework'.  Stop.
make: *** No rule to make target `lib', needed by `framework'.  Stop.
building JSON framework
xcodebuild -create-xcframework -output ../dist/frameworks/JSON.xcframework -library build_result/iphone/libjson.a -library build_result/simulator/libjson.a
xcframework successfully written out to: /Users/adil/Work/j2objc/dist/frameworks/JSON.xcframework
```
It seems that only the final command in the for loop in the [build_subset_frameworks.sh][4] script succeeds, i.e. the command to build the JSON framework.

[...]

PR #2166, opened 2023-09-12
Fixed the mk files in the 'subset_frameworks' directory

adil-hussain-84:
Fixed the mk files in the [subset_frameworks][2] directory by following the pattern in the [subset_frameworks/json_framework.mk][1] file. Fixes [#2165][3].

[1]: https://github.com/google/j2objc/blob/master/jre_emul/subset_frameworks/json_framework.mk
[2]: https://github.com/google/j2objc/blob/master/jre_emul/subset_frameworks
[3]: https://github.com/google/j2objc/issues/2165


adil-hussain-84:
The CI Workflow Run for this pull request failed for a reason that is unrelated to the code change. See Issue #2167 for the details.

jre_emul/build_subset_frameworks.sh
```diff
@@ -17,7 +17,7 @@
 # 1. cd to the j2objc/jre_emul directory
 # 2. run "./build_subset_frameworks.sh"
 #
-# To delete these frameworks, run "./build_subset_frameworks clean".
+# To delete these frameworks, run "./build_subset_frameworks.sh clean".
 
 for mk in \
   android_util_framework.mk \
```
tomball, line 23: Please remove the android_util_framework.mk line, and delete that file. The android_util library isn't built any, and there's an error reported when this script runs this make command.

tomball (changes requested):
Looks good over all, just one error to be addressed and it's ready.

jre_emul/build_subset_frameworks.sh
adil-hussain-84: Done.

tomball (approved):
LGTM

adil-hussain-84:
Argh, this branch's CI build will fail because it doesn't contain the change which just went into master to update the CI to always build with Java 11. I'll decline this MR, rebase my changes on top of master and create a new pull request.

tomball:
FYI, there's no need to do that in the future. Github-only files, like `.github/workflows/main.yml`, get directly merged after review, but anything that touches the project's files have to get pulled into our internal repo first and reviewed there, before getting pushed out. We have internal CIs that are more thorough than Github's (that's why we hadn't noticed that the Github CI was broken), so this wouldn't have been an issue. Regardless, thanks for your contributions!

tomball:
Argh, you're right -- the failing CI means that this approved MR isn't being forwarded internally.

Issue #2169, opened 2023-09-13
Cannot use JRE.xcframework with Xcode 15 Release Candidate due to code signing issue


mungler:
Hi,

We've been using JRE.xcframework for a while now, and its been working great. We note the imminent release of macOS Sonoma and Xcode 15, so decided to try building in the Xcode 15 RC. However, our app now refuses to build, with the error: 

"<app path>/frameworks/JRE.xcframework:1:1 The signature of “JRE.xcframework” cannot be verified", and when clicked, offers to move it to the trash, with a further message:

"Validation Error: bundle format unrecognized, invalid, or unsuitable"

I've tried:

- changing JRE.xcframework to 'Embed & Sign' in the 'Frameworks, Libraries and Embedded Content' section of the 'General' tab of the build settings. 
- code signing JRE.xcframework manually via the `codesign` tool, but it gives me the same 'unrecognized, invalid, or unsuitable' message.
- code signing the individual libjre_emul.a files inside the various platform-specific subfolders

None of these seem to help.

I'm wondering if there's something missing from the generated xcframework bundles - i've seen reference to a Resources folder, is there perhaps some file that needs to live there?

Anyone with a better handle on this stuff able to help?

mungler:
Closing in favour of #2101 - sorry, failed to notice this has been logged already. This will prevent us upgrading to Sonoma when its released later this month until its resolved.

bryan1anderson:
You included a couple steps I hadn't tried yet. Thanks for this. Also have run into the same problem..

litstrong:
Hi mungler@, may I ask where you see the references to the Resources folder?

mungler:
Re: "Resources", i'll try to find out, sorry, I just noticed in passing when I spent most of yesterday googling on the subject. Will update here if/when I find anything concrete.

Here's the thing that makes me think the package is malformed somehow - even reading the signing info results in an error. Compare:

```
rory@Zaphod ~> touch foo
rory@Zaphod ~> codesign -dv foo
foo: code object is not signed at all

rory@Zaphod ~ [1]> codesign --timestamp -s "Apple Development: Rory Sinclair (99******HA)" foo
rory@Zaphod ~> codesign -dv foo
Executable=/Users/rory/foo
Identifier=foo
Format=generic
CodeDirectory v=20200 size=131 flags=0x0(none) hashes=0+2 location=embedded
Signature size=9172
Timestamp=14 Sep 2023 at 09:24:29
Info.plist=not bound
[...]

Issue #2101, opened 2023-06-06
[Xcode 15-Beta] - The signature of “JRE.xcframework” cannot be verified


ferencIOS:
Hi all.

New WWDC new problems!

When I try to run the app that contains the JRE.xcframework, the new Xcode shows me this error:
`Codesign verification failed`

They must surely have introduced some new flag that is not known.

(I'm not alone: https://developer.apple.com/forums/thread/730955)

Thanks for your help!

tomball:
Are you building j2objc as part of your app build, or using a pre-built release? We're going to release a new version once the JRE update is finished, as the current version is likely out-of-date.

ferencIOS:
Basically I followed what I had answered some time ago [here](https://github.com/google/j2objc/issues/1933#issuecomment-1338972728)!

Then I integrated the xcframework into the project.

They probably added some new flag for:
```
xcodebuild clean archive ...
```

In fact it works with Xcode 14 and it doesn't work with Xcode 15

ferencIOS:
Following the Apple WWDC:

https://developer.apple.com/documentation/Xcode/verifying-the-origin-of-your-xcframeworks
or
https://developer.apple.com/videos/play/wwdc2023/10061/

which simplifying
```
xcodebuild -create-xcframework [...] -output <YOUR.xcframework>
codesign --timestamp -v --sign "<YOUR CERTIFICATE (ABCXYZ)>" "<YOUR.xcframework>"
```

dsanghan:
Yup seeing the same issue. Trying to manually code sign it causes the error:
```
JRE.xcframework: bundle format unrecognized, invalid, or unsuitable
```

dsanghan:
@tomball Any updates on this?

tomball:
No, I don't have a personal certificate, don't have any access to a Google-wide one, and will be retiring soon. @litstrong (j2objc's project lead), do you have any update?

[One of you can contribute a fix](https://developers.google.com/j2objc/guides/how-to-contribute), as the `xcodebuild -create-xcframework` command is in the short [gen_xcframework.sh](https://github.com/google/j2objc/blob/master/scripts/gen_xcframework.sh) script. Since the certificate is personal and shouldn't be checked into any project, I suggest supporting a shell environment variable that defines it, then update the script to reference that variable.

litstrong:
Ack, sorry for my late response.

Just start to pick this up, let me take a look and will respond by next week.

litstrong:
[...]

PR #2175, opened 2023-09-14
Change XCFramework artefact format to keep Xcode 15 happy

mungler:
When building XCFramework artefacts, don't create Versions folder with Headers symlink; create real Headers folder instead

### DIFF
--- a/make/framework.mk
+++ b/make/framework.mk
@@ -99,13 +99,11 @@ $(FRAMEWORK_DIR): lib $(FRAMEWORK_HEADER) $(MODULE_MAP) | $(DIST_FRAMEWORK_DIR)
 	@mkdir -p $(FRAMEWORK_DIR)
 	@$(J2OBJC_ROOT)/scripts/gen_xcframework.sh $(FRAMEWORK_DIR) \
 		$(shell $(J2OBJC_ROOT)/scripts/list_framework_libraries.sh $(STATIC_LIBRARY_NAME))
-	@mkdir -p $(FRAMEWORK_DIR)/Versions/A/Headers
-	@/bin/ln -sfh A $(FRAMEWORK_DIR)/Versions/Current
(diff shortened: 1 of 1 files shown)


mungler:
@litstrong @tomball FYI this fixes issues #2169 and #2101

litstrong (approved):
Thanks for your investigation and fixing this!

make/framework.mk
```diff
@@ -99,13 +99,11 @@ $(FRAMEWORK_DIR): lib $(FRAMEWORK_HEADER) $(MODULE_MAP) | $(DIST_FRAMEWORK_DIR)
 	@mkdir -p $(FRAMEWORK_DIR)
 	@$(J2OBJC_ROOT)/scripts/gen_xcframework.sh $(FRAMEWORK_DIR) \
 		$(shell $(J2OBJC_ROOT)/scripts/list_framework_libraries.sh $(STATIC_LIBRARY_NAME))
-	@mkdir -p $(FRAMEWORK_DIR)/Versions/A/Headers
-	@/bin/ln -sfh A $(FRAMEWORK_DIR)/Versions/Current
-	@/bin/ln -sfh Versions/Current/Headers $(FRAMEWORK_DIR)/Headers
+	@mkdir -p $(FRAMEWORK_DIR)/Headers
 	@tar cf - -C $(STATIC_HEADERS_DIR) $(FRAMEWORK_HEADERS:$(STATIC_HEADERS_DIR)/%=%) \
-		| tar xfp - -C $(FRAMEWORK_DIR)/Versions/A/Headers
```
litstrong, line 106: Shall we guard the change by the Xcode version so that it won't be broken when running for old Xcode version?

mungler: Only if we're sure the newer-style framework doesn't work on older Xcode, I guess. Unfortunately I only have one Mac these days and after resolving this have upgraded to Sonoma RC, so am unable to run older Xcode versions now :)

PR #2181, opened 2023-09-19
Now calling the `xcodebuild -create-xcframework` command with the `-headers` option

### LINKED ISSUE #2180: Headers information is missing from the XCFrameworks' "Info.plist" files
The XCFrameworks which are built right now place the `Headers` directory at the root-level of the XCFramework and omit mention of the headers in the XCFramework's `Info.plist` file.

We need to update the [gen_xcframework.sh][1] script so that it calls the `xcodebuild -create-xcframework` command with the `-headers` option as described in the [Generate the XCFramework bundle][2] Apple Developer guide. This will create a `Headers` directory and a `HeadersPath` entry in the `Info.plist` file for each architecture in the XCFramework.

I will create a Pull Request with this change shortly.

[1]: https://github.com/google/j2objc/blob/master/scripts/gen_xcframework.sh
[2]: https://developer.apple.com/documentation/xcode/creating-a-multi-platform-binary-framework-bundle#Generate-the-XCFramework-bundle

adil-hussain-84:
Updated the `make/framework.mk` and `scripts/gen_xcframework.sh` files so that the `xcodebuild -create-xcframework` command is called with the `-headers` option. Fixes #2180.

### DIFF
--- a/make/framework.mk
+++ b/make/framework.mk
@@ -96,14 +96,15 @@ framework:: lib $(FRAMEWORK_DIR) resources
 # Create an xcframework from all appletv, iphone, maccatalyst, macosx, simulator and watchos libs.
 $(FRAMEWORK_DIR): lib $(FRAMEWORK_HEADER) $(MODULE_MAP) | $(DIST_FRAMEWORK_DIR)
 	@echo building $(FRAMEWORK_NAME) framework
-	@mkdir -p $(FRAMEWORK_DIR)
-	@$(J2OBJC_ROOT)/scripts/gen_xcframework.sh $(FRAMEWORK_DIR) \
-		$(shell $(J2OBJC_ROOT)/scripts/list_framework_libraries.sh $(STATIC_LIBRARY_NAME))
+	@rm -rf ${FRAMEWORK_DIR}
 	@mkdir -p $(FRAMEWORK_DIR)/Headers
 	@tar cf - -C $(STATIC_HEADERS_DIR) $(FRAMEWORK_HEADERS:$(STATIC_HEADERS_DIR)/%=%) \
 		| tar xfp - -C $(FRAMEWORK_DIR)/Headers
 	@install -m 0644 $(FRAMEWORK_HEADER) $(FRAMEWORK_DIR)/Headers
 	@install -m 0644 $(MODULE_MAP) $(FRAMEWORK_DIR)/Headers/
+	@$(J2OBJC_ROOT)/scripts/gen_xcframework.sh $(FRAMEWORK_DIR) \
+		$(shell $(J2OBJC_ROOT)/scripts/list_framework_libraries.sh $(STATIC_LIBRARY_NAME))
+	@rm -rf ${FRAMEWORK_DIR}/Headers
 	@touch $@
 
 # Creates a framework "master" header file that includes all the framework's header files.
--- a/scripts/gen_xcframework.sh
+++ b/scripts/gen_xcframework.sh
@@ -12,12 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-# Build XCFramework framework from set of single-arch static libraries.
+# Build an XCFramework from a set of single-architecture static libraries.
 # xcodebuild -create-xcframework needs all libraries to be included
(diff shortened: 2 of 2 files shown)


make/framework.mk
```diff
@@ -96,14 +96,15 @@ framework:: lib $(FRAMEWORK_DIR) resources
...
-	@mkdir -p $(FRAMEWORK_DIR)
-	@$(J2OBJC_ROOT)/scripts/gen_xcframework.sh $(FRAMEWORK_DIR) \
-		$(shell $(J2OBJC_ROOT)/scripts/list_framework_libraries.sh $(STATIC_LIBRARY_NAME))
+	@rm -rf ${FRAMEWORK_DIR}
 	@mkdir -p $(FRAMEWORK_DIR)/Headers
 	@tar cf - -C $(STATIC_HEADERS_DIR) $(FRAMEWORK_HEADERS:$(STATIC_HEADERS_DIR)/%=%) \
 		| tar xfp - -C $(FRAMEWORK_DIR)/Headers
 	@install -m 0644 $(FRAMEWORK_HEADER) $(FRAMEWORK_DIR)/Headers
 	@install -m 0644 $(MODULE_MAP) $(FRAMEWORK_DIR)/Headers/
+	@$(J2OBJC_ROOT)/scripts/gen_xcframework.sh $(FRAMEWORK_DIR) \
```
adil-hussain-84, line 105: Calling the `gen_xcframework.sh` script is moved to later in the make task because it now requires that the `$(FRAMEWORK_DIR)/Headers` directory exists and that it contains all its content.

scripts/gen_xcframework.sh
```diff
@@ -27,14 +27,18 @@ fi
...
+if [ ! -d "${FRAMEWORK_HEADERS_DIR}" ]; then
+  echo "$FRAMEWORK_HEADERS_DIR should exist but doesn't."
+fi
 
 CMD="xcodebuild -create-xcframework -output "${FRAMEWORK_DIR}
 
 while test ${#} -gt 0
 do
   CMD=${CMD}" -library "$1
+  CMD=${CMD}" -headers "$FRAMEWORK_HEADERS_DIR
```
adil-hussain-84, line 41: See the [Generate the XCFramework bundle](https://developer.apple.com/documentation/xcode/creating-a-multi-platform-binary-framework-bundle#Generate-the-XCFramework-bundle) Apple Developer guide for an explanation of the `-headers` option.

mungler:
Nice work @adil-hussain-84 - this does indeed produce valid XCFramework artefacts which seem to comply with Apple's specification. The /usr/bin/codesign tool is still happy that its a valid framework, and thus Xcode 15 will compile projects using the framework without issue.

👍 for merge

mungler (approved):

mungler:
@tomball FYI I have approved this PR but still see the message "At least 1 approving review is required by reviewers with write access." - does someone else need to approve, or will this be picked up automatically?

tomball (approved):
Thanks for the clear comments describing what's done in this PR. No need to guess!

tomball:
Yes, someone from  Github's Google org needs to approve, which I just did.
Instead of merging this PR, the next process step will cause the PR to be
staged internally as a change request, requiring another engineer to
approve, and then it gets automatically merged (Google has to be
super-careful about how changes are integrated into its source base). This
should happen automatically, but I'm always a little amazed when it works.
:-)

mungler:
Haha sounds good @tomball 👌

tomball:
[Copybara](https://github.com/google/copybara) is our repo-syncing service. We're good here, and thanks to you both for your contribution and help!

mungler:
I did very little other than try out @adil-hussain-84 ’s changes, but you’re most welcome 👌

adil-hussain-84:
@tomball @mungler: Thank you both for reviewing, approving and merging 👍👍 Sadly, it's looking like this not the last pull request from me related to XCFrameworks 😭 A number of the XCFrameworks that are being built have the same module name (i.e. `JRE`) in their `module.modulemap` file. The implication of this is that two modules with the same cannot be linked into the same target. I see that their is an open Issue and Pull Request related to this already. I'll post in there and draw attention to it.

PR #2182, opened 2023-09-20
Fixed the module names in the `module.modulemap` files

### LINKED ISSUE #2111: JSON.xcframework contains wrong module.modulemap
Hi,

I made a build of j2objc - the current master branch. It looks like there is an issue in Makefile when creating JSON.xcframework. The framework contains .modulemap from JRE framework:

```
module JRE {
  umbrella header "JRE.h"

  export *
  module * { export * }
}
```

instead of

```
module JSON {
  umbrella header "JSON.h"

  export *
  module * { export * }
}
```

The problem is in framework.mk file, the path to modulemap $(BUILD_DIR)/module.modulemap for JRE and JSON is the same, so that it is used for JSON.xcframework as well.

adil-hussain-84:
Changed the `make/framework.mk` file so that the name of the module in the generated `module.modulemap` file corresponds to the module being built. Fixes #2111.

### DIFF
--- a/make/framework.mk
+++ b/make/framework.mk
@@ -57,8 +57,8 @@ FRAMEWORK_PUBLIC_HEADERS = $(FRAMEWORK_HEADERS)
 endif
 
 FRAMEWORK_DIR = $(DIST_FRAMEWORK_DIR)/$(FRAMEWORK_NAME).xcframework
-FRAMEWORK_HEADER = $(BUILD_DIR)/$(FRAMEWORK_NAME).h
-MODULE_MAP = $(BUILD_DIR)/module.modulemap
+FRAMEWORK_HEADER = $(BUILD_DIR)/$(FRAMEWORK_NAME)/$(FRAMEWORK_NAME).h
+MODULE_MAP = $(BUILD_DIR)/$(FRAMEWORK_NAME)/module.modulemap
 
 FRAMEWORK_RESOURCES_DIR = $(FRAMEWORK_DIR)/Versions/A/Resources
 RESOURCE_FILES = $(FRAMEWORK_RESOURCE_FILES:%=$(FRAMEWORK_RESOURCES_DIR)/%)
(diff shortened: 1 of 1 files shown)


make/framework.mk
```diff
@@ -57,8 +57,8 @@ FRAMEWORK_PUBLIC_HEADERS = $(FRAMEWORK_HEADERS)
 endif
 
 FRAMEWORK_DIR = $(DIST_FRAMEWORK_DIR)/$(FRAMEWORK_NAME).xcframework
-FRAMEWORK_HEADER = $(BUILD_DIR)/$(FRAMEWORK_NAME).h
-MODULE_MAP = $(BUILD_DIR)/module.modulemap
+FRAMEWORK_HEADER = $(BUILD_DIR)/$(FRAMEWORK_NAME)/$(FRAMEWORK_NAME).h
+MODULE_MAP = $(BUILD_DIR)/$(FRAMEWORK_NAME)/module.modulemap
```
adil-hussain-84, line 61: I've changed the `MODULE_MAP` path so that it's different for each framework that is built by this make file. Without this change, the frameworks that are built as part of the `build_subset_frameworks.sh` script all use the `module.modulemap` file that corresponds to the `JRE` framework and hence all have a module name of `JRE`. That is bad because Xcode forbids linking to two modules of the same name.

make/framework.mk
```diff
@@ -128,6 +129,7 @@ test_warnings: $(FRAMEWORK_HEADER)
 	@rm $(FRAMEWORK_HEADER:%.h=%.o)
 
 $(MODULE_MAP):
+	@mkdir -p $$(dirname $@)
 	@echo "module" $(FRAMEWORK_NAME) "{" > $(MODULE_MAP)
```
adil-hussain-84, line 133: I was tempted to change all occurrences of `$(MODULE_MAP)` in the `$(MODULE_MAP)` target's commands to `$@` but I decided against it. I'll leave that for a follow-up pull request in order to keep this one focused on the change at hand.

make/framework.mk
```diff
@@ -57,8 +57,8 @@ FRAMEWORK_PUBLIC_HEADERS = $(FRAMEWORK_HEADERS)
 endif
 
 FRAMEWORK_DIR = $(DIST_FRAMEWORK_DIR)/$(FRAMEWORK_NAME).xcframework
-FRAMEWORK_HEADER = $(BUILD_DIR)/$(FRAMEWORK_NAME).h
-MODULE_MAP = $(BUILD_DIR)/module.modulemap
+FRAMEWORK_HEADER = $(BUILD_DIR)/$(FRAMEWORK_NAME)/$(FRAMEWORK_NAME).h
```
adil-hussain-84, line 60: It wasn't strictly necessary to move the `FRAMEWORK_HEADER` path from `$(BUILD_DIR)` to `$(BUILD_DIR)/$(FRAMEWORK_NAME)`. I made this change so that all build files for a given framework (i.e. both the framework header and the module map) are in the same place.

tomball (approved):
LGTM, thanks!

tomball (commented):
LGTM

PR #2184, opened 2023-09-21
Add missing headers for JRE_Core XCFramework

ale-gen:
Having added JRE_Core.xcframework to the project (built from the latest commit) and imported <java/lang/Integer.h> to the Cocoa Touch Class'.m file, the build fails with error: `'java/nio/channels/InterruptibleChannel.h` file not found`. Theses changes fix this issue 🚀

### DIFF
--- a/jre_emul/jre_sources.mk
+++ b/jre_emul/jre_sources.mk
@@ -306,6 +306,7 @@ JAVA_PUBLIC_SOURCES_CORE = \
   java/nio/channels/ClosedChannelException.java \
   java/nio/channels/FileChannel.java \
   java/nio/channels/GatheringByteChannel.java \
+  java/nio/channels/InterruptibleChannel.java \
   java/nio/channels/ReadableByteChannel.java \
   java/nio/channels/ScatteringByteChannel.java \
   java/nio/channels/SeekableByteChannel.java \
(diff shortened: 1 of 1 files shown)


tomball:
Thanks for wanting to contribute to the project! However, those headers are not missing, they're in JRE_channels.xcframework, which is one of a subset frameworks of JRE.xcframework. You can find which classes are in which subset in [jre_classes.mk](https://github.com/google/j2objc/blob/master/jre_emul/jre_sources.mk).

If you don't want to deal with JRE subsets, JRE.xcframework has headers for all public JRE classes. Since the reason for using JRE subsets instead is to reduce app size, it is generally easiest to use the JRE framework during initial development (when which JRE classes used may not be known), and then switch to subset frameworks when the project is close to completion.

ale-gen:
Thanks for the response! However, I'm still facing up with errors in my sample app. Basically, I hoped that I could add the JRE_Core XCFramework to an Xcode target without needing linking to any other of the J2ObjC XCFramework. I believed that JRE_Core XCFramework should be self-contained. Besides when I add mentioned JRE_Channels etc. XCFrameworks I get "multiple commands produce 'module.modulemap'" error 🤔

tomball:
JRE_Core is self-contained; it's defined by the minimum set of JRE classes needed to link a "hello, world" source as a macOS command-line app. Each app has its own "core" set of classes needed to be fully linked, and so most need JRE subset libraries to augment this minimum core set. We keep core as small as possible because every class we move from a subset increases the size of all apps that otherwise wouldn't link with that subset.

You're real issue here is #2185. @adil-hussain-84 is working on that and hopes to have a fix next week. I therefore recommend that you either wait for his fix, or use JRE.xcframework (no subsets). The Apple docs on XCFramework are very incomplete, so if you have any insights on how this can be fixed, please comment on #2185 and/or work with @adil-hussain-84.

tomball:
Ah, it looks like you already have in #2189. Thanks!

Issue #2185, opened 2023-09-21
Encountering "multiple commands produce 'module.modulemap'" error when linking a target with any two of the J2ObjC XCFrameworks


adil-hussain-84:
I've built J2ObjC from the [dd82e40][1] commit (dated September 20, 2023).

When I link an Xcode application or framework target with any two of the XCFrameworks produced by J2ObjC (e.g. `JRE_Core` and `JSON`), I'm encountering the following build error:

```none
Multiple commands produce '<Build Location>/Products/Debug-iphonesimulator/include/module.modulemap'
```

My suspicion is that the `module.modulemap` file within the XCFrameworks is in the wrong location. Right now it's placed in the `<LibraryIdentifier>/Headers` directory. My feeling (yet to be proven through experimentation) is that it should be in the `<LibraryIdentifier>/Modules` directory.

I plan on spending more time in investigating and fixing this but I won't get time until Monday. I'm putting it out there in case anybody else is affected by it or wants to pick it up in the meanwhile.

[1]: https://github.com/google/j2objc/commit/dd82e40e7271684f573a887d82b5a347c0b3a341

adil-hussain-84:
I ran some experiments with @ale-gen today and we've come to the conclusion that the `module.modulemap` file does not make sense for the XCFrameworks produced by the J2ObjC build scripts given that the XCFrameworks are packaging up libraries and not frameworks. (It may have made sense in the past but the behaviour of Xcode 14 leads us to believe that it is now redundant. I will explain below.)

An XCFramework which is built with the `xcodebuild -create-xcframework -framework <path> [-framework <path>...] -output <path>` command and thus packages up a set of frameworks has the following directory structure:

```none
MyFramework.xcframework (directory)
  |-- <LibraryIdentifier> (directory) (e.g. `ios-arm64`)
    |-- MyFramework.framework (directory)
      |-- Info.plist (file)
      |-- MyFramework (binary file)
      |-- Headers (directory)
      |-- Modules (directory)
        |-- module.modulemap (file)
        |-- MyFramework.swiftmodule (directory)
```

On the other hand, an XCFramework which is built with the `xcodebuild -create-xcframework -library <path> [-headers <path>] [-library <path> [-headers <path>]...] -output <path>` command and thus packages up a set of libraries has the following directory structure:

```none
MyLibrary.xcframework (directory)
  |-- <LibraryIdentifier> (directory) (e.g. `ios-arm64`)
    |-- Info.plist (file)
[...]

PR #2189, opened 2023-09-22
Fix "multiple commands produce 'module.modulemap'" error

ale-gen:
Based on [that](https://github.com/google/j2objc/pull/2184) PR and its response, I've tried to link to other mentioned J2ObjC's XCFrameworks, but it caused "multiple commands produce 'module.modulemap'" error.  Also I hoped that I could add the JRE_Core XCFramework to an Xcode target without needing linking to any other of the J2ObjC XCFramework. I believed that JRE_Core XCFramework should be self-contained. Because of that I created [sample app](https://github.com/ale-gen/ModulemapErrorSampleApp) to show my problem. 

When I linked to another XCFramework produced by J2ObjC, I got the same situation like described in issue [#2185](https://github.com/google/j2objc/issues/2185). With changes added to this PR, I've noticed that in my case I don't have to link to another XCFramework, but if I do this, I also won't get modulemap error! It seems that currently modulemap file is located in wrong place.

### DIFF
--- a/make/framework.mk
+++ b/make/framework.mk
@@ -98,10 +98,11 @@ $(FRAMEWORK_DIR): lib $(FRAMEWORK_HEADER) $(MODULE_MAP) | $(DIST_FRAMEWORK_DIR)
 	@echo building $(FRAMEWORK_NAME) framework
 	@rm -rf ${FRAMEWORK_DIR}
 	@mkdir -p $(FRAMEWORK_DIR)/Headers
+	@mkdir -p $(FRAMEWORK_DIR)/Modules
 	@tar cf - -C $(STATIC_HEADERS_DIR) $(FRAMEWORK_HEADERS:$(STATIC_HEADERS_DIR)/%=%) \
 		| tar xfp - -C $(FRAMEWORK_DIR)/Headers
 	@install -m 0644 $(FRAMEWORK_HEADER) $(FRAMEWORK_DIR)/Headers
-	@install -m 0644 $(MODULE_MAP) $(FRAMEWORK_DIR)/Headers/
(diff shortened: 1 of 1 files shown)


tomball:
@adil-hussain-84, would you please review and approve this change? I'll need to as well, but you have the build environment to (hopefully) quickly verify this fix is correct.

@ale-gen, would you please accept the Google CLA? Click on the Details link above in the cla/google error above.

tomball:
@ale-gen: thanks for this change, I was able to build the Hello example with multiple JRE subset frameworks successfully. Where did you find current documentation on XCFramework layout?

ale-gen:
> @ale-gen: thanks for this change, I was able to build the Hello example with multiple JRE subset frameworks successfully. Where did you find current documentation on XCFramework layout?

Unfortunately I didn't find official documentation, so I looked for some .xcframework examples, and then I came upon [this XCFramework's contents explanation](https://github.com/bielikb/xcframeworks#contents-of-xcframework)

adil-hussain-84:
@tomball: Sure. I'll experiment with this change today on my machine and determine whether it solves Issue #2185.

Just to let you know, @ale-gen is working with me on the same Xcode framework project. We're trying to change our Xcode framework target from its current setup of linking the `libjre_core.a`, `libjson.a` and `libjsr305.a` static libraries to a new setup of linking the `JRE_Core.xcframework`, `JSON.xcframework` and `JSR305.xcframework` frameworks. We're doing this as a stepping stone on the way to solving a larger problem: An app which links to our framework (which links to the above-mentioned J2ObjC static libraries) now happens to link to another framework which also links to some J2ObjC libraries 😭 According to Murphy's Law, it was only a matter of time!

PR #2193, opened 2023-09-25
Removed the `module.modulemap` file from the XCFrameworks

adil-hussain-84:
Changed the `framework.mk` file so that it no longer makes the `module.modulemap` file as part of its execution. In summary, the `module.modulemap` file does not serve a purpose within XCFrameworks which bundle up a set of static libraries. See the discussion in Issue #2185 for more detailed rationale. This change fixes issue #2185.

### DIFF
--- a/make/framework.mk
+++ b/make/framework.mk
@@ -58,7 +58,6 @@ endif
 
 FRAMEWORK_DIR = $(DIST_FRAMEWORK_DIR)/$(FRAMEWORK_NAME).xcframework
 FRAMEWORK_HEADER = $(BUILD_DIR)/$(FRAMEWORK_NAME)/$(FRAMEWORK_NAME).h
-MODULE_MAP = $(BUILD_DIR)/$(FRAMEWORK_NAME)/module.modulemap
 
 FRAMEWORK_RESOURCES_DIR = $(FRAMEWORK_DIR)/Versions/A/Resources
 RESOURCE_FILES = $(FRAMEWORK_RESOURCE_FILES:%=$(FRAMEWORK_RESOURCES_DIR)/%)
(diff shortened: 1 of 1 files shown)


tomball (approved):
LGTM, thanks!

tomball:
Merge will be done in shortly by the [copybara service](https://github.com/google/copybara).

make/framework.mk
```diff
@@ -128,15 +126,6 @@ test_warnings: $(FRAMEWORK_HEADER)
 		-fno-objc-arc $@
 	@rm $(FRAMEWORK_HEADER:%.h=%.o)
 
-$(MODULE_MAP):
```
litstrong, line 131: The module map looks useful for Swift interop purpose. If we remove them here, does that mean we need to import it somewhere again when necessary?

litstrong (approved):

make/framework.mk
adil-hussain-84: I'm running some experiments to get to a somewhat definitive answer to this question. I'll get back to you by the end of the week.

adil-hussain-84: @litstrong: I ran a number of experiments and I couldn't figure out a way to define a static library which consists of Objective-C classes only such that it can be imported as a module into another module. The only way that I could get the library's headers to become visible to Swift classes in the importing module was to add the library's headers to the importing module's Objective-C Bridging Header.

I compared the build outputs produced by a static library which consists of Objective-C classes only against the build outputs produced by a static library which contains of one or more Swift classes. I'll summarise what I saw for each.

In the case of a static library which consists of Objective-C classes only, the build outputs were as follows:

```none
+-- libMyLibrary.a (file)
\-- include (directory)
  \-- MyLibrary (directory of headers)
```

In the case of a static library which contains one or more Swift classes, the build outputs were as follows:

```none
+-- libMyLibrary.a (file)
\-- MyLibrary.swiftmodule (directory)
```

The `swiftmodule` directory seems to me to be the key to allowing a library to be imported as a module in the importing module's Swift code. If a library does not produce a `swiftmodule` directory, the way to use the library's Objective-C classes in the importing module's Swift code is to add the library's headers to the importing module's Objective-C Bridging Header.

PR #2302, opened 2024-03-07
feat: support jpms in `annotations` module

sgammon:
## Summary

This PR adds a `module-info.java` definition to the `annotations` package, and adjusts the build to produce a [MRJAR](https://openjdk.org/jeps/238). The result is a JAR artifact which (1) remains safely usable on JDK 1.8, and (2) helps enable modular Java builds for consumers of the annotations.

Downstream, Guava depends on these annotations, and [many artifacts depend on Guava](https://mvnrepository.com/artifact/com.google.guava/guava). As a result, many projects that don't otherwise use J2ObjC or ship for iOS rely on J2ObjC adopting a module definition like this one.

I've made the minimum suite of changes possible here to maintain both the `make` and `maven`-based builds. The two produce nearly identical JARs, save for the Maven Compiler Plugin stamping the JAR manifest differently with its own message. I've cleaned and performed a full rebuild of the project to ensure the JARs are properly usable.

For now, I will build and issue the JAR artifacts at the version `2.8-jpms`, and make them available for downstream testing. I think releasing the annotations under a version postfix like that might help projects opt-in before this ships for real.

## Changelog

- feat: add `module-info.java` to `annotations`
- chore: wire up build support for building `java9` root in a multi-release jar
- chore: apply java9 jpms changes to both maven and makefile build

Relates to: google/guava#2970


sgammon:
cc / @cpovirk

annotations/.gitignore
```diff
@@ -0,0 +1 @@
+/target
```
sgammon, line 1: When building `annotations` with `mvn package`, `target` shows up as an untracked Git directory.

annotations/Makefile
```diff
@@ -25,17 +25,25 @@ DIST_JAR = $(DIST_JAR_DIR)/$(JAR_NAME)
 
 CLASSES_DIR = $(BUILD_DIR)/classes
 JAVA_SOURCE_DIR = src/main/java
+JAVA9_CLASSES_DIR = $(BUILD_DIR)/classes-java9
+JPMS_CLASS = $(JAVA9_CLASSES_DIR)/module-info.class
```
sgammon, line 29: Classes for Java 9 are built together with the `module-info.class`. This is necessary because the `annotations` should export all annotation classes, and `javac` wants to be able to see those during the module build.

We don't want to include the duplicate Java 9 bytecode classes, though, just the `module-info.class`. So they are _built_ together in this dedicated directory, but only the `module-info.class` makes it into the final MR JAR.

annotations/Makefile
```diff
@@ -25,17 +25,25 @@ DIST_JAR = $(DIST_JAR_DIR)/$(JAR_NAME)
...
+	@echo Building j2objc annotations JPMS module
+	@$(JAVAC) -sourcepath $(JAVA_SOURCE_DIR) -encoding UTF-8 --release 9 -d $(JAVA9_CLASSES_DIR) \
+		-nowarn $(SOURCE_JAVA9_MODULE)
+
+$(JAR): $(SOURCE_JAVA_FULL) | $(BUILD_DIR) $(CLASSES_DIR) $(JAVA9_CLASSES_DIR) $(JPMS_CLASS)
 	@echo Building j2objc annotations
 	@$(JAVAC) -encoding UTF-8 -d $(CLASSES_DIR) -source 1.8 -target 1.8 \
 	    -nowarn $^
-	@jar cf $(JAR) -C $(CLASSES_DIR) .
+	@jar cf $(JAR) -C $(CLASSES_DIR) . --release 9 -C $(JAVA9_CLASSES_DIR) .
```
sgammon: Makefile build logic; the JPMS class is built first, as described above, and then included in the `jar cf` command at the end.

annotations/Makefile
```diff
@@ -25,17 +25,26 @@ DIST_JAR = $(DIST_JAR_DIR)/$(JAR_NAME)
...
+	@$(JAVAC) -sourcepath $(JAVA_SOURCE_DIR) -encoding UTF-8 --release 9 -d $(JAVA9_CLASSES_DIR) \
+		-nowarn $(SOURCE_JAVA9_MODULE)
+	@rm -fr $(JAVA9_CLASSES_DIR)/com
+
+$(JAR): $(SOURCE_JAVA_FULL) | $(BUILD_DIR) $(CLASSES_DIR) $(JAVA9_CLASSES_DIR) $(JPMS_CLASS)
 	@echo Building j2objc annotations
 	@$(JAVAC) -encoding UTF-8 -d $(CLASSES_DIR) -source 1.8 -target 1.8 \
 	    -nowarn $^
-	@jar cf $(JAR) -C $(CLASSES_DIR) .
+	@jar cf $(JAR) -C $(CLASSES_DIR) . --release 9 -C $(JAVA9_CLASSES_DIR) .
```
sgammon, line 47: Main Makefile build logic for the MRJAR/Java9 change. JPMS module is built first so that `module-info.class` is available by the time `jar cf ...` runs, since `javac` and `jar` are in one block.

The main class root (`com/`) doesn't exist in `classes-java9` by the time this runs, so just the `module-info.class` is included in the Java 9 root. This results in the following JAR layout:

```
j2objc-annotations.jar/
├─ com/
│  ├─ google/
├─ META-INF/
│  ├─ versions/
│  │  ├─ 9/
│  │  │  ├─ module-info.class
│  ├─ MANIFEST.MF
```

Full JAR as built by Maven:
Screenshot 2024-03-07 at 3 48 51 PM

Full JAR as built by Makefile:
Screenshot 2024-03-07 at 3 49 35 PM

annotations/pom.xml
```diff
@@ -58,6 +58,49 @@
...
+        <executions>
+          <execution>
+            <id>default-compile</id>
+            <configuration>
+              <source>1.8</source>
+              <target>1.8</target>
+              <excludes>
+                <exclude>module-info.java</exclude>
+              </excludes>
+            </configuration>
```
sgammon, line 74: Compiles for `1.8` at the root of the JAR.

annotations/pom.xml
```diff
@@ -58,6 +58,49 @@
...
+          <execution>
+            <id>compile-java9</id>
+            <phase>compile</phase>
+            <goals>
+              <goal>compile</goal>
+            </goals>
+            <configuration>
+              <release>9</release>
+              <multiReleaseOutput>true</multiReleaseOutput>
+            </configuration>
```
sgammon, line 85: Compiles at `--release=9` and produces MRJAR output.

annotations/pom.xml
```diff
@@ -58,6 +58,49 @@
...
+        <version>3.3.0</version>
+        <configuration>
+          <archive>
+            <manifestEntries>
+              <Multi-Release>true</Multi-Release>
+            </manifestEntries>
+          </archive>
+          <excludes>
+            <exclude>META-INF/versions/9/com/**</exclude>
+          </excludes>
```
sgammon, line 101: Filters out duplicate annotation classes.

annotations/src/main/java/module-info.java
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+open module com.google.j2objc.annotations {
+  requires java.base;
+  exports com.google.j2objc.annotations;
+}
```
sgammon, line 20: I decided to express this as an `open module ...` so that annotations can be introspected safely.

sgammon:
`make -j$(nproc) && make test_all` passes, tested in the environment shown below. If there are any notable cases to cover, I'm happy to test locally, or amend CI.

```
CPU: M-series Mac (aarch64)
OS: macOS Sonoma `14.3.1`, Java 17 (Azul Zulu).

> java -version
openjdk version "11.0.19" 2023-04-18 LTS
OpenJDK Runtime Environment Zulu11.64+19-CA (build 11.0.19+7-LTS)
OpenJDK 64-Bit Server VM Zulu11.64+19-CA (build 11.0.19+7-LTS, mixed mode)
```

annotations/.gitignore
tomball: Please delete this file, as the real file is in the parent directory (j2objc/.gitignore).

annotations/Makefile
tomball: SGTM

annotations/Makefile
tomball: Thanks for the thorough explanation!

annotations/src/main/java/module-info.java
tomball: SGTM

tomball (approved):
Thank you very much for this contribution, and all your notes regarding it. I'll work with the Guava team when this is committed so their next release uses this version.

sgammon:
@tomball Thank you for the quick response! And for all of your hard work on J2ObjC. 😄  I'll keep an eye on this in case you need any changes from me.

cpovirk:
Thanks, @sgammon and @tomball!

Before your PR yesterday, I had been playing with the alternative of replacing Guava's `j2objc-annotations` dependency with local copies of the annotations in each Guava package that needed them. That was looking OK up until I realized that it would mean something like 20 annotations (since J2ObjC provides 4 different annotations, most of them used from various Guava packages). We _could_ do that, but it feels more bloated and clumsy (especially if we need to introduce more copies in more packages over time).

So it will be great to have `j2objc-annotations` ready for modules. (No doubt some users would _still_ prefer for us to remove the dependency entirely :) But I am much more concerned with the module issue than I am with the "I don't want Maven to pull in a jar of 10 annotations by default" issue.)

sgammon:
@cpovirk If 10 annotation classes are still too much, I can offer a PR which links users to Proguard ;)

sgammon:
Thank you both for the quick review and merge. Made my day! 🥳

tomball:
Right, that's one reason none of these annotations have
RetentionPolicy.RUNTIME -- they have nothing to do with code execution,
only with compiler (j2objc)
generation.

sgammon:
@tomball Happy meals were exciting to me when you started coding in Java. So I will defer to you, this is why

Issue #2501, opened 2025-06-11
InputStreamReader memory leak


braandl:
Hello,

I have a strange problem with the java.io.input.StreamReader,
it holds a cyclic reference between the java.io.Reader's `lock` and the member `sd` which is the StreamDecoder. All usages are directly from within the `jre_emul`.

Attached find a Screenshot from the iOS memory graph attached. Maybe someone has an idea for me, or is this a weak spot in the current jre_emul?

![Image](https://github.com/user-attachments/assets/34da0b02-3df1-4ae3-a9b8-fd5523adb14d)

tomball:
Fixed in latest commit.

PR #2508, opened 2025-06-17
Setting Null-Reference on for originalBufferObject on MemoryRef

braandl:
This is another memory-leak that happens within my setup using the java.nio.* implementation. Just like shown in https://github.com/google/j2objc/issues/2501 it creates cyclic reference between the `DirectByteBuffer` and `MemoryRef`. 

Since the mechanism was created to hold a reference to the original Buffer, after the "live" reference might has been deallocated, this seems not to be a working mechanism in Objective-C.

This change sets the originalBufferObject to always null (like on the other constructor already) to avoid this behavior.
Therefore now the MemoryRef loses its functionality of keeping the DirectByteBuffer's ref. I could not find any negative effects tough, since I am convinced, this can not happen in the transpiled framework anyways.


braandl:
I did after I created this. Somehow the bot does not recognize it, but the website states my state is fine.

tomball (approved):
Looks good, thanks for the contribution!

tomball:
Your PR is merged, thanks for your contribution (and patience!).

PR #2604, opened 2025-11-02
Fixed testSkipTreeKindImport() translator test, updates build to JDK 21 minimum.

tomball:
Fixed testSkipTreeKindImport() so it tests that the expected javac error is returned.


kevinoconnor7:
FYI @tomball piper is the Source of Truth for j2objc so copybara needs to be the one to actually merge (which it will do after it has been submitted in piper).

I'll manually merge this in piper to get things back into a good state.

Issue #2565, opened 2025-09-20
Add java.util.Collection.toArray(IntFunction<T[]>), introduced in JDK 11


AsheraCordova:
When we try to compile this file:

https://github.com/AsheraCordova/AndroidJ/blob/main/src/main/java/r/android/util/MapCollections.java

The source is translated to objc.

But gives the following error:

- (IOSObjectArray *)toArrayWithJavaUtilFunctionIntFunction:(id<JavaUtilFunctionIntFunction>)arg0 {
  return JavaUtilCollection_toArrayWithJavaUtilFunctionIntFunction_(self, arg0);
}

JavaUtilCollection_toArrayWithJavaUtilFunctionIntFunction_ is not found.. I hato manually remove this method.

I think that:

    default <T> T[] toArray(IntFunction<T[]> generator) {
        return toArray(generator.apply(0));
    }

has not been implemented in java.util.Collection but the translator is generating the code

CristianLopezR:
I think the issue has been resolved

PR #2611, opened 2025-11-04
Issue #2565: restored `java.util.Collection.toArray(IntFunction<T[]>)`

tomball:
Restored `java.util.Collection.toArray(IntFunction<T[]>)` and implementations in `java.util.Collections`. 
These methods were commented out as they were new to JDK 11, while j2objc was still based on JDK 1.8.


rluble:
FYI, it is failing kokoro (it does not have Java 21 installed but we are working on it)

rluble:
Tests are failing due to `toArray(null)` becoming ambiguous in the jsr166 collection tests where there are a few classes with tests like this one:

```java
   public void testToArray_NullArray() {
        final Collection q = emptyCollection();
        try {
            q.toArray(null);
            shouldThrow();
        } catch (NullPointerException success) {}
    }
```

These are quite old tests and the updated versions have a cast `(Object[])` to resolve the ambiguity. Could you add these fixes to the pr?

tomball:
I don't know, but if this isn't merged tomorrow morning, then maybe you'll need to reapprove in the code review panel so copybara pulls it in.

rluble (approved):

PR #2618, opened 2025-11-06
Fixed translator's Java sources list in public build.

tomball:
`translate/SwitchConstructRewriter.java` replaced `translate/SwitchCaseRewriter.java`, so this updates the Make build's sources list.


rluble:
Ahh. I didn't see that. I also renamed the corresponding test.

rluble (approved):

PR #2670, opened 2026-02-10
Do not increase pointer after last enum constant

zbynek:
The last occurence of `ptr += objSize` in enum declaration is redundant and produces warnings with `xcodebuild ... analyze`.

This PR makes sure this extra assignment is skipped.


zbynek:
@rluble sorry to bother you, but I didn't get any feedback on this since February, so I'm wondering if this is something Google would be interested in merging.

rluble:
Sorry for the late reply. Let me have a look.

zbynek:
@rluble gentle ping

PR #2671, opened 2026-02-10
Do not treat Mongolian vowel separator as a whitespace

zbynek:
As of Java 11 `isWhitespace` for `\u180e` is false.

This aligns the Java emulation with GWT 2.12+
https://github.com/gwtproject/gwt/issues/9973#issuecomment-2214224281


tomball:
The j2objc version of Character.java is a fork of [Android's libcore version](https://android.googlesource.com/platform/libcore/+/master/ojluni/src/main/java/java/lang/Character.java), not GWT's. Although there's nothing wrong with this change, it would be better if the whole `isWhitespace()` method were copied, to maintain j2objc compatibility with Android's JRE.

